PR #14, opened 2017-01-10
Added syntax highlighting for code

jacksarick:


mobitar:
Hey, thanks for submitting this. It's tempting, given how simple the changes are, to merge this right away. However, I'm haunted by the mission of this project: to survive for a really long time. In order to accomplish that, this project needs to be as lightweight and nimble as possible. Every dependency we add takes ten years off its life, and every one I've already added has been done so in mourning. 

I do however understand the need for this app to be what a lot of people need it to be, so I'm thinking of ways to add these sorts of things in as extensions. I've already laid the groundwork for a data [extension system](https://standardnotes.org/extensions), with [Dropbox sync](https://standardnotes.org/extensions/dropbox) being the first available. 

As far as UI extensions, I'm coming close towards a solution, but not quite there yet. Once I figure out a way to allow custom functionality, like Markdown or code highlighting, we'll want to start beginning adding things like this. Because in the year 2030, maybe there'll be a newer Markdown alternative, and everyone starts requesting that. Rather than denying what the market wants, it's better to make this extensible by default. If you have any ideas on solutions, feel free to throw them my way.

Btw, love your website. Minimal and clean. My eyes pop when I see stuff like that.

PR #35, opened 2017-01-18
Sort tags by name

fdhoff:
Sorts tags by their lowercase title as new tags are added. 

Todo: render newly added tags according to the sort order.


mobitar:
Nice use of `sortedIndexBy`! Didn't know lodash had that.

fdhoff:
Thanks!  Now we both know hahah

PR #42, opened 2017-01-20
Only handle tab key inside of editor

levlaz:
This refactors the controller a bit to insert 4 spaces when
a tab key is inserted inside of an editor, but in all other parts
of the app allow the user to use the tab key normally.

fix
https://github.com/standardnotes/desktop/issues/23

### DIFF
--- a/app/assets/javascripts/app/frontend/controllers/editor.js
+++ b/app/assets/javascripts/app/frontend/controllers/editor.js
@@ -60,7 +60,7 @@ angular.module('app.frontend')
               }
           }
         };
-	
+
         window.addEventListener('keydown', handler);
         var element = document.getElementById("note-text-editor");
         element.addEventListener('keydown', handleTab);
--- a/app/assets/javascripts/app/frontend/controllers/editor.js
+++ b/app/assets/javascripts/app/frontend/controllers/editor.js
@@ -19,8 +19,7 @@ angular.module('app.frontend')
          * Insert 4 spaces when a tab key is pressed, 
          * only used when inside of the text editor.
          */
-        var handleTab = function (event ) {
-
+        var handleTab = function (event) {
           if (event.which == 9) {
             event.preventDefault();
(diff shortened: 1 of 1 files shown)


mobitar:
I don't think the `editor` reference will transfer from the link function into the controller function. I get a `ReferenceError: editor is not defined` when I press enter from the title (which focuses the editor).

Instead you'll want to set it to `ctrl.editor = ...` in the link function, and access it by `this.editor` in the controller function.

levlaz:
Nice catch, sorry again. 

Ok, so using `this.editor` did not seem to work either because it was complaining that it was undefined. In addition ctrl.editor was also not defined. 

I just grabbed the element again and left the `controller` piece alone. 

This might be ok, I can imagine a couple of other handy keyboard shortcuts inside of the textarea such as `ctrl+a = home` and `ctrl+e = end` which can reuse the separate function that only handles the keydowns inside of the textarea.

mobitar:
Looks good 👍 Btw the reason the editor was undefined in the controller was probably because `this` was undefined because you had to bind it to the timeout method:

```
setTimeout(function(){

}.bind(this), delay)
```

Just for future reference.

app/assets/javascripts/app/frontend/controllers/editor.js
```diff
@@ -15,16 +15,22 @@ angular.module('app.frontend')
...
 
+        /**
+         * Insert 4 spaces when a tab key is pressed, 
+         * only used when inside of the text editor.
+         */
+        var handleTab = function (event) {
+          if (event.which == 9) {
+            event.preventDefault();
+            var start = event.target.selectionStart;
+            var end = event.target.selectionEnd;
```
levlaz: Bah, I found another bug. 

If you try add a tab when you have more text after it. (i.e. in the middle of the note) it jumps to the end of the document. I think its because of this hack. Will look into this.

PR #43, opened 2017-01-20
change undefined titles to empty string

fdhoff:
When a user creates a new tag with empty title, ensure the title's value is set to empty string `''` instead of  `undefined`.


mobitar:
Hmm..I don't think the right place to do this would be modelManager addItems. Because not all items have titles. An Item could be anything. In our case yes both tags and notes have title, but this isn't guaranteed in the future.

I think a better place would be to do this when the item is actually created, and set defaults there.

But why this change in the first place? Is there something bad happening when it's undefined?

Issue #50, opened 2017-01-23
[bug] - Tab key inside of a note goes to the end of document 


levlaz:
See [here](https://github.com/standardnotes/web/pull/42/files#r97437269) for details. 

Sorry.. I have no idea why handling tabs is so complicated. :)

levlaz:
@mobitar just reporting this here in case someone else notices this issue, I will send a PR with a fix.

mobitar:
Oh wow didn't even realize that, but you're right.

levlaz:
Yeah, I use the app every day now and ran into this. The good news is I can only blame myself. :D

djhalliday:
Thanks for looking into this @levlaz, I've started using the app every day and this is definitely one of the most maddening bugs!

levlaz:
@djhalliday my pleasure! I totally agree that this was driving me crazy as well. I am also sad that I shipped this bug. :) 

I submitted a PR that fixes this. Thanks so much for bearing with it.

mobitar:
I also noticed that I couldn't shift + tab from the textarea to go back to the title. Do you think that's an easy fix?

levlaz:
Yeah for sure, we can add a condition to not fire this off when shift is pressed.

levlaz:
@mobitar I updated the PR to handle Shift + Tab with the browser default.

PR #52, opened 2017-01-24
Reset the cursor

levlaz:
Once tab key is pressed, place the cursor 4 spaces away from
the location where the tab key was pressed.

This handles the case of when a tab key is pressed in the middle of
an existing note and the cursor was placed at the end of the document.

fix https://github.com/standardnotes/web/issues/50


mobitar:
Works great, thanks @levlaz

PR #54, opened 2017-01-25
Set preview mode as default

### LINKED ISSUE #31: Option to set Markdown Preview as default open mode.
Have an option to set preview mode as the default after a note is created. Currently, it defaults to edit mode. I use notebooks as a reference and read from my notes more frequently than I write.

When I open a note that uses image embed or formatting I only see the markdown text until switching to preview mode:

markdown

Having the preview mode as default would be ideal.

amtsh:
## Current scenario
Default editor mode is 'edit' for notes.

## Fix done
After a note is created, the default mode of the editor is now set to preview.
For new note, the default mode is edit.

fix #31

### DIFF
--- a/app/assets/javascripts/app/frontend/controllers/editor.js
+++ b/app/assets/javascripts/app/frontend/controllers/editor.js
@@ -91,7 +91,8 @@ angular.module('app.frontend')
   .controller('EditorCtrl', function ($sce, $timeout, apiController, markdownRenderer, $rootScope, extensionManager) {
 
     this.setNote = function(note, oldNote) {
-      this.editorMode = 'edit';
+      var isNewNote = note.dummy
+      this.editorMode =  isNewNote ? 'edit' : 'preview';
       this.showExtensions = false;
       this.showMenu = false;
       this.loadTagsString();


mobitar:
I guess my issue with this is, if you're presenting your note to someone else, why would you want to view it as Markdown? Typically when I'm composing Markdown documents, as an author, I like to preview them. But it would be strange if, say, upon opening a Markdown file, Atom started it in preview mode. Wouldn't seem to make much sense. 

Can you describe your use case for this? I don't want to create an additional step in between a user editing their note.

gabrielcossette:
Either you want to read or edit your notes. To read them, Markdown Preview is better of course.

The default should be chosen depending on which use case is thought to be more popular among users.

Whatever is chosen, as long as there's an option/config to change it, users will be happy :)

gabrielcossette:
But personally, I prefer Markdown Preview by default because it looks better when I want to give them a quick glance, even before editing one.

mobitar:
When in doubt, we have to go with what's standard. Standard is a plain text editor. Everything you add on top of that is custom. Markdown is custom. Favoring Markdown = not standard, less sustainable. It's definitely a hard decision, but I'm hoping it's for the best.

amtsh:
I agree with you 'Favoring Markdown = not standard, less sustainable'

How about 'having an option/config to change it'
We can have a 'Settings' menu in the header and on/off element to change the default behavior.

mobitar:
Settings mean more code. And once we have a settings panel, that's the end of Standard Notes as we know it :) We'll just keep adding things to it. Let's hold off on this and see if we can find a better solution.

mobitar:
Closing as preview mode has been removed from the app, in favor of editor extensions.

Issue #53, opened 2017-01-25
When offline, add "Offline" in bold red to bottom right corner


mobitar:
When you're signed in, but for some reason unable to reach the server, we should add this text so the user knows changes are being saved only to local DB.

PR #57, opened 2017-01-28
Show 'Offline' in bottom right of header when offline

amtsh:
For a signed in user, the 'offline' text will appear in the header for 2 cases:
1. The user clicks the 'Refresh' button in the header and the server is unreachable.
2. The user is typing the note in the editor, the autosave fails as the server is unreachable, hence the note status shows '(Offline) — All changes saved'.

**This PR has the case 1.**

For case 2, I am stuck and need help.
- While editing the note, if remote server is unreachable, the note status says [this](https://github.com/standardnotes/web/blob/master/app/assets/javascripts/app/frontend/controllers/editor.js#L158)
I am wondering, how I can set the isOffline boolean in HeaderCtrl from EditorCtrl.
- I tried to add a 'offlineManager' service that will have isOffline flag and then inject this service in HeaderCtrl and EditorCtrl.
But if EditorCtrl changes the flag when offline, then the header have to know it without checking the flag explicitly.

Please suggest what should be the right way in this case.

### DIFF
--- a/app/assets/javascripts/app/frontend/controllers/header.js
+++ b/app/assets/javascripts/app/frontend/controllers/header.js
@@ -129,8 +129,10 @@ angular.module('app.frontend')
           this.isRefreshing = false;
         }.bind(this), 200)
         if(response && response.error) {
+          this.isOffline = true
           alert("There was an error syncing. Please try again. If all else fails, log out and log back in.");
         } else {
+          this.isOffline = false
           this.syncUpdated();
         }
       }.bind(this));
--- a/app/assets/stylesheets/app/_header.scss
+++ b/app/assets/stylesheets/app/_header.scss
@@ -65,6 +65,10 @@
        color: #515263;
      }
 
+     &.offline {
+       color : red;
+     }
+
      .panel {
        position: absolute;
        right: 0px;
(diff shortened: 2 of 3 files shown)


mobitar:
Wow, you read my mind. I have the exact same changes in this branch that I'll be merging sometime today: https://github.com/standardnotes/web/pull/58/files

Long story short, I spent about a week (a very long week) working on adding support for multiple sync providers in the app, before realizing yesterday this functionality would be better handled by the server (attaching a Dropbox account to the server account, etc). 

So I undid all those changes, but since I had made many good refactors, I kept those around. Which is why that branch is called "sync_providers_destruction". Anyway, these changes will:

1. Display Offline in lower right corner if user is not signed in
2. If there is an error syncing:
 - Display "Sync error" in red where it currently says "All changes saved".
 - Make the "Account" button in lower left corner red.
 - Display the error message in the account panel.
3. Display sync status in the account panel. i.e "syncing 5/10 items".

I appreciate you looking out for this and recognizing the need for it.

mobitar:
Whoops, I just realized I created this issue: https://github.com/standardnotes/web/issues/53.

I should have communicated that I was working on it. Totally forgot about it. Sorry about that.

amtsh:
Wow, that's a very huge refactor.

No problem, I didnt spend much time on this. :)
I will close this pull request.

PR #67, opened 2017-02-08
After deleting a note, it should select the first note on top

### LINKED ISSUE #63: After deleting a note, it should select the first note on top
Currently the editor just disappears

amtsh:
**Current scenario :** 
After deleting a note, the editor just disappears

**Fix :**
After deleting a note, select first note in the current notes list.

fix #63


mobitar:
Thanks @amtsh, will test this this week hopefully. Also, what's your email address? Wanted to send you a message.

amtsh:
Cool.
Email : shinde13amit@gmail.com

mobitar:
So the current paradigm is that all the controls (tags, notes, editor) talk with the home controller to perform any actions. They don't broadcast or talk with one another. Do you think it's possible to accomplish this same thing using existing hooks, with the home controller coordinating everything?

PR #81, opened 2017-02-26
remove restangular as dependency, use xmlhttprequest

mobitar:


chuckdries:
Just saw your message about this merge on Slack. Even though this is already merged, my life is basically just a long chain of sleep and XMLHttpRequest Promise resolutions so I figured I'd weigh in.

Your implementation of HttpManager looks good from a vanilla js standpoint. The only thing I'd worry about is how you go about handling params when you're POSTing them. I'm not familiar with the framework you're replacing, but some services expect POST data as a url encoded string instead of a JSON string, though from poking around the project it looks like everyone speaks JSON around here so it's probably a moot point.

I'm assuming you went with the callback injection pattern you did because you wanted a drop in replacement for whatever you're replacing. That's obviously fine and works alright, but you don't provide a way to access to your request objects outside of your manager class. Perhaps limiting yourself to using the get and post helper methods you defined in the class is a design choice. I've found, however, that returning the XMLHttpRequest after creating it ends up being very useful and adds a lot of flexibility down the road. You can add additional event handlers after firing the request or cache the request object itself which is sometimes more convenient than caching the data you extract from it. The biggest advantage of this is freeing you from having to put all of the logic you need to deal with particularly large requests into massive monolithic request-specific callbacks.

Another example: I've had instances where I have to make n requests for n pieces of similar information, in which case the fact that you can't pass arguments or use .bind() on callback functions a real pain. I get around this by wrapping the returned XMLHttpRequest object in a lightweight object that contains a reference to the piece of information that called it and defining the callback function as a method of that object. In this way, you can store an array of requests and when your request comes back your generic callback can know exactly what it's dealing with.

Lately, I've been wrapping my requests in Promises instead of my home grown requestWrapper objects. The great thing about Promises is how they elegantly handle event binding and chaining. 

If I have a promise `myPromise`, there are two main functions I'll use to interact with it: `myPromise.then()` and `myPromise.catch()`. Both of these methods take a callback as an input and return the modified Promise with the new event listener attached. That means you can write nice code like this:
```javascript
sendRequest({
        url: requrl,
        params: {
            key: devkey
        },
        method: 'GET'
    })
    .then(cleanData)
    .then(renderList)
    .catch((err) => {
        console.error("error occured!");
        console.error(err);
    });
```
`.then()` and `.catch()` both pass their data to their callback functions just as you'd expect, but they actually chain the data (in this case, the responseText from the XMLHttpRequest wrapped within the promise... more on this in a moment) in the same way that they chain the rest of the Promise. When I call `.then(cleanData)`, cleanData is passed the response text, `JSON.parse()`s it, modifies it a bit for my specific application, then returns it. The cool thing is, now the Promise that the current `then()` returns holds the return value of `cleanData()` where it used to hold the raw responseText. In a normal callback function setup, I'd either have to call `renderList()` within `cleanData()` every time I use `cleanData()`, implement some logic in `cleanData()` to decide whether to call `renderList()` (which would certainly not be the right place for such logic), or, more likely, define an anonymous function that exists purely to call those two other functions. Not only do I find that annoying, but it also rebinds `this` so you have to call `.bind()` all over the place. 

With this Promise, `renderList` will be passed the output of `cleanData`. I can choose to call them separately from each other on a per request basis whenever I want in a clean and easy format. Of course, they're not totally independent; `renderList()` will be expecting cleaned data, but I can easily swap out the action I take after cleaning the data, even at runtime ;).

Here's how I wrap the promise in the current (somewhat small) project I'm working on, though it could easily be updated to work within a factory/manager object like you're using.

```javascript
function sendRequest(options) {
        return new Promise(function(resolve, reject) {
            var req = new XMLHttpRequest();
            var params = options.params;
            var id = options.id;
            //stringify params if not already a string. URL encoded or JSON encoded
            if (params && typeof params === 'object') {
                if (options.method == "GET" || options.paramStyle == "QS") { //specify querystring or sending in GET request
                    params = Object.keys(params).map((key) => (
                        encodeURIComponent(key) + '=' + encodeURIComponent(params[key])
                    )).join('&');
                } else { //JSON by default
                    params = JSON.stringify(params);
                }
            }
            if (options.method == "GET") {
                options.url = options.url + "?" + params;
            }
            req.open(options.method, options.url)
            req.onload = function() {
                //to be clear, this refers to the XMLHttpRequest object.
                if (this.status == 200 && this.readyState == 4) {
                    //the Promise constructor takes one argument: the activator function. It passes this activator (or it might be called something else?) a resolve and reject function, which, when called, will trigger .then() and .catch() and pass their callback functions whatever data you pass them here.
                    //I pass an ID value associated with the Promise to keep track of the requests. This specific webapp interfaces with a 10,700 line google spreadsheet via the sheets API. Keeping track of requests is crucial as it fetches resources based on data stored in rows of the sheet.
                    resolve(this.response, id);
                } else { //TODO: errored promises do not always contain a statustext. Further testing required
                    reject({
                        status: this.status,
                        statusText: this.statusText
                    });
                }
            }
            req.onerror = function() {
                reject({
                    status: this.status,
                    statusText: req.statusText
                });
            }
            if (options.headers) {
                Object.keys(options.headers).forEach((key) => {
                    req.setRequestHeader(key, options.headers[key]);
                });
            }
            req.send(params);
        });
    }
```

I can't wait until I have more time to start contributing to this project. Fantastic work.

mobitar:
Thanks for your review, wish I had someone to look at it while building it ;) I come from a Swift/Objective-C background and have only picked up JavaScript in the last couple years, so most implementations I go with are the most basic way of doing something. For the most part, it turns out to be a good thing since it keep things simple and not over-engineered. But I can see how promises can be useful here. The library I'm replacing here is Restangular, which had a nice API similar to what you mentioned:

```
request.post()
.then(function(response){})
.catch(function(error){})
```

What we have now seems to be working, though I wouldn't mind a cleaner implementation if we have the luxury of time for it (and as long as it keeps things lightweight).

PR #101, opened 2017-03-30
Add Docker support

tommyku:


Gemfile
```diff
@@ -21,6 +21,8 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 # Used for 'respond_to' feature
 gem 'responders', '~> 2.0'
 
+gem 'tzinfo-data'
```
mobitar: @tommyku what is the rationale for this?

tommyku: Had to install this or the server doesn't start

mobitar: That's strange. What was the issue?

tommyku: The output looks like this if I attempt to start the Rails server without `tzinfo-data`, any idea how to fix it?

image

tommyku: I think I got it, quotes 
> On most Unix-based systems (e.g. Linux), TZInfo is able to use the system zoneinfo directory as a source of data.

from (tzinfo doc)[https://github.com/tzinfo/tzinfo/wiki/Resolving-TZInfo::DataSourceNotFound-Errors]

The fact that I used alpine linux instead or something bigger (say, Ubuntu) have caused this

PR #109, opened 2017-04-27
Clear New Theme URL after submit

aaronraimist:
This commit clears the "New Theme URL" after submitting the form in the same way that it is done for adding a new editor via URL.


mobitar:
Good catch, thanks.

PR #110, opened 2017-04-27
Update Gemfile to allow for Ruby 2.4.1

aaronraimist:
What version of Ruby should be used when developing Standard Notes web? Either that version should be documented in the README or this pull request should be accepted which updates a couple of dependencies in the Gemfile to allow Standard Notes to be built with the current stable Ruby version 2.4.1.

I haven't fully tested every feature using these new dependencies but it seems to work. More testing may be required.


mobitar:
Hmm I'll have to test this one carefully. I use Ruby 2.3.1 in production.

PR #119, opened 2017-06-12
Use password field type for passwords

tcyrus:
The Change Password and Import Password fields should not be plain text fields.


mobitar:
Thanks!

PR #178, opened 2018-02-04
Remove dead code for installing new editors

bnjbvr:
Thanks for doing Standard Notes! I noticed `handleEditorLink` isn't defined and the proper way of setting editors is a Component with area editor-editor, so I think this code is dead and can be removed.


mobitar:
Good news—globalExtensionsMenu.js is completely gone in the packages branch. All of the extension logic has been moved to the extensions-manager extension: https://github.com/sn-extensions/extensions-manager

bnjbvr:
Indeed, I was lacking some context with respect to the update here. Thanks!

PR #184, opened 2018-02-15
Fixes #183: In tag views, only display other tags on notes;

### LINKED ISSUE #183: In note list view, all tags shown if a note has several tags
I've got a note that has tags `A` and `B`. When I click on the tag `A` in the tags list (so as to filter only notes with tag `A`), all the notes don't show the tag `A` in their list of tags, except for this one note that has several tags. Adding a screenshot to make it more explicit:

screenshot-2018-2-15 standard notes a private and secure notes app

I'd expect the selected note to only show `#geek`, not `#associatif #geek`.

Does it make sense? Thanks!

bnjbvr:
I didn't see a lot of ES6 code, so I assumed babel wasn't used, so the code I introduced is very conservative with ES features, let me know if you'd prefer something else.

Can you have a look please, @mobitar ? Thanks!

### DIFF
--- a/app/assets/javascripts/app/controllers/notes.js
+++ b/app/assets/javascripts/app/controllers/notes.js
@@ -271,4 +271,14 @@ angular.module('app')
       return note.tags && note.tags.length > 1;
     }
 
+    this.getTagsString = function(note) {
+      var tags;
+      if(this.tag.all) {
+        tags = note.tags;
+      } else {
+        tags = note.tags.filter(tag => tag.uuid !== this.tag.uuid);
+      }
+      return Tag.arrayToDisplayString(tags);
+    }
+
   });
--- a/app/assets/javascripts/app/models/app/note.js
+++ b/app/assets/javascripts/app/models/app/note.js
@@ -108,7 +108,7 @@ class Note extends Item {
   }
 
   toJSON() {
-    return {uuid: this.uuid}
+    return {uuid: this.uuid};
   }
 
   get content_type() {
--- a/app/assets/templates/notes.html.haml
+++ b/app/assets/templates/notes.html.haml
@@ -55,7 +55,7 @@
             %strong.medium Archived
 
           .tags-string{"ng-if" => "ctrl.shouldShowTags(note)"}
-            .faded {{note.savedTagsString || note.tagsString()}}
+            .faded {{ ctrl.getTagsString(note) }}
 
           .name{"ng-if" => "note.title"}
             {{note.title}}


mobitar:
Looks great! It does indeed use babel and ES6. You can just run `grunt` to compile.

Only two changes that I think might be helpful:
1. Filter by tag.uuid instead of title, since you can have multiple tags with the same name. This probably doesn't matter in most cases, but in the case of a sync conflict where a tag is duplicated, it might. Better to be on the safe side.
2. Rename `showTags` to a return-based name, like `getTagsString` or something similar. IMO `showTags` sounds like a verb, so implies it has side-effects. This could just be a personal preference, but better to be consistent.

Thanks for the PR!

bnjbvr:
Thanks for you feedback, it all makes sense! Code has changed around there, but I think I've found my way. Can you have another look, please? Thanks!

mobitar:
Hmm..https://github.com/standardnotes/web/blob/master/app/assets/javascripts/app/controllers/editor.js#L334

bnjbvr:
Aouch, I even missed two (a second one in this file). Is there an easy/quick way to retrieve the current tag from the list, in the editor?

mobitar:
Get the currently selected tag from the editor controller? Not that I know of. Best to keep that separate. What are you trying to do?

bnjbvr:
Trying to reuse the same code that's in controllers/notes.js at the moment, to avoid code duplication. But maybe we can simply have `Note.getTagStrings` take a supplementary facultative parameter that is the tag to ignore? (not sure what the other location I'm touching does affect, will investigate more)

mobitar:
The notes.js controller has a reference to the currently selected tag, and is also the one that displays the tags string. Why not keep everything in there?

bnjbvr:
My intent was to avoid code duplication with the one location you've found there: https://github.com/standardnotes/web/blob/master/app/assets/javascripts/app/controllers/editor.js#L334

But maybe there's no need to filter at all on this line? (= what I'd like to investigate)

mobitar:
The reason we filter is to avoid a redundant sync request. This function is called when the tags input loses focus, which can happen just as you're tabbing from input to input. If that check is not there, a sync will be triggered, marking the note dirty, and updating it's `updated_at` property, even though nothing was changed.

bnjbvr:
Thanks for your answer! I think there's been a misunderstanding: my question about "filtering" was to know whether we needed to filter the current selected tag (as in selected from the tag list) in the code location I've pinpointed. I now think I got it: this part is the editor where we can actually enter tags, so we don't want to hide any tags.

I've updated my patch so that only the list of notes is affected by the tag filtering. This looks slightly more clean in my opinion (we could cache the filtered tags list as well, but I don't know if it's worth doing).

Does it look better now? I tested it a bit more and it works nicely.

bnjbvr:
Right, do I cache it as `note.savedTagsWithoutEditorTagString` then? Is there a code location where I'd need to clear it, when a note gets dirtied?

mobitar:
Sorry we didn't get far on this one. Appreciate your help nonetheless!

PR #186, opened 2018-02-22
fix(app/assets): Make some buttons focusable

sorin-davidoi:
Divs are not focusable by default, and need a considerable amount of JavaScript to make they behave similar to the native button element. This switches buttons implemented as divs to the native button element (styles are updated accordingly to maintain the same visual style). Link tags without href are also not focusable - for most of these cases the button element should be used, but that would involve a more significant refactoring, so I have opted to adding a dummy href.

While enabling keyboard navigation, this also addresses the fact that screen reader users had no idea that these buttons existed.

### DIFF
--- a/app/assets/stylesheets/app/_main.scss
+++ b/app/assets/stylesheets/app/_main.scss
@@ -170,6 +170,7 @@ $footer-height: 32px;
          border-radius: 4px;
          font-weight: normal;
          text-align: center;
+         border: none;
 
          &:hover {
            background-color: darken($button-bg, 5%);
--- a/app/assets/stylesheets/app/_menus.scss
+++ b/app/assets/stylesheets/app/_menus.scss
@@ -1,6 +1,8 @@
 .app-bar {
   .item {
     position: relative;
+    background: inherit;
+    border: none;
   }
 }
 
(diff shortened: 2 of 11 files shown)


mobitar:
Hey @sorin-davidoi, thanks for the PRs! These are far-reaching changes, so before I test everything locally myself, I just want to make sure you have. Were you able to get the app environment set up locally and confirm that none of these changes affect the UI or interactions?

sorin-davidoi:
Yes, I've tested - the only visual change that I could notice it that the buttons have the default outline when they are focused.

mobitar:
Ok great. I'm planning on revisiting this repo sometime next week, and will review these changes then. Thanks!

sorin-davidoi:
@mobitar Anything else I can do here (and #187)?

mobitar:
Hi Sorin, sorry for the delay. We're currently trying to sort out our licensing for this repo, and until then, it would be best to not merge any PRs (given we'd have to relicense your changes; something I'm newly learning about). Ongoing discussion here: https://github.com/standardnotes/web/issues/189

sorin-davidoi:
@mobitar Has the license issue been settled? The issue is still opened but I see that pull requests from other contributors have been accepted.

mobitar:
It has; however, your PR changes fundamental behavior across many areas of the app, which makes it difficult to test. PRs that are focused on a single area are much less risky to merge in. Also, there haven't been any functionality-focused PRs that have been merged in the last year, from what I can see. The ones that you do see merged are all my own PRs that I create to review big changes.

sorin-davidoi:
It would be preferable to open one PR per file?

PR #223, opened 2018-08-08
Run app locally without having to know about grunt

DeveloperDavo:
* Also updated grunt and npm rearranged the packages into alphabetical order
* It might worth reminding the user to install rails beforehand. (I have never used rails before and I had to look this up)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ This repo contains the core code used in the web app, as well as the Electron-ba
 1. `bundle install`
 1. `npm install`
 1. `bundle exec rake bower:install`
-1. `grunt`
+1. `npm run build`
 1. `rails s`
 
 Open your browser to http://localhost:3000.
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,8 +10,8 @@
       "integrity": "sha512-3Sp6WZZ/lXl+nTDoGpGWHEpTnnC6X5fnkolYZR6nwIfzbxxvA8utPWe1gCt7i0m9uVGsSz2IS8K8mJ7HmlduMg==",
       "dev": true,
       "requires": {
-        "jsonparse": "1.3.1",
-        "through": "2.3.8"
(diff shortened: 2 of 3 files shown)


mobitar:
Looks good! Will have to test the grunt version upgrade though. Also, if you're simplifying the build with `npm run build`, couldn't a bunch of other steps be put into here as well?

"build" : "bundle install && npm install && bundle exec rake bower:install && grunt"

Haven't tested this, but sounds like it would be possible. Only problem would be if there is an issue with one of those commands, might be a little harder for the average user to debug.

DeveloperDavo:
So the disadvantage I see is that you don't always need to run those other commands as frequently. I see it more of a set up than a build step but it's up to you of course :)

mobitar:
Hmm ok..I'm still not fully sold this is a positive improvement. It's not negative either. Just sort of neutral. Might it not be good for people to understand grunt to know they need to compile their code every time they change something? And also there's `grunt watch`, for when you want it to compile your code automatically as you change it.

DeveloperDavo:
So the main reason I did it was because the README says to run `grunt` and I thought having a global installation was required. So I installed it as a dependency, without realising it was already there, and thought it would be a good idea to abstract away from it with the npm script. Abstracting away from it has the advantage of swapping out the build tool without having to update documentation and I believe it also convention. A compromise might be to write `npx grunt` instead of `grunt` as one of the steps. 

I would have created a separate script for watching as well :D. 

Anyway, I won't be offended if you reject the pull request.

mobitar:
Makes sense. I think I'm on board. It would make it a lot easier to test for me if you didn't upgrade grunt. Just not sure what side effects the new version will have. This will also clear the merge conflicts. Can you update the PR just to update just the package.json scripts section and readme?

DeveloperDavo:
Ok. I hope that is what you meant :)

PR #229, opened 2018-08-31
Update Dockerfile

oldjamey:
Also added instructions for building and deploying Docker image, and added docker directory with custom entrypoint.  I have tested and confirmed that this works for deploying to localhost.  There are basic instructions in the Dockerfile for making this externally-facing using Nginx and LetsEncrypt SSL.


mobitar:
Excellent, thank you!

PR #230, opened 2018-09-03
Update Dockerfile Comments

oldjamey:
The "run" instructions had a different image name from the "build" instructions.  This update just makes them work correctly together.


mobitar:
👍👍

PR #235, opened 2018-09-05
Docker/prod config

phx:
Commented-out portions of Dockerfile that can be uncommented for use in production.  Also added a comment to config/environments/production.rb to change the value of a parameter for configurations that include HTTP-only localhost service accessed by upstream reverse proxy terminating SSL.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -1,6 +1,7 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
-
+  
+  # Set the value below to false if you are terminating SSL upstream and proxying to HTTP:
   config.force_ssl = true
 
   # Code is not reloaded between requests.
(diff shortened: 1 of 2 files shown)


config/environments/production.rb
```diff
@@ -1,6 +1,7 @@
 Rails.application.configure do
   # Settings specified here will take precedence over those in config/application.rb.
-
+  
+  # Set the value below to false if you are terminating SSL upstream and proxying to HTTP:
```
mobitar: Regarding this comment, I think it's too specific to be permanently here. Better to create a separate issue and people with this issue can search these keywords.

oldjamey: I agree. It was late when I was working on this.  After further review, this comment needs to go into the Dockerfile under the commented out production configuration.  I will update this.

oldjamey: See new pull request #236 which removes the production.rb comment from the equation, but adds to the Dockerfile comments, because this is normally the way Docker apps are deployed in production.

PR #257, opened 2018-12-03
When in the middle of a list, continue the list when Enter key is pressed

Arrow7000:
I.e. when you press the Enter key when your current line starts with, for example, `- ` (a dash followed by a space) the next line will automatically start with `- ` as well, making it more convenient to write lists.

Currently acceptable list item markers are the characters `-`, `·` and `•`; but more could be added.

I have closely followed the existing coding style, especially basing my `handleListEnter` helper function on the existing `handleTab`.

### DIFF
--- a/app/assets/javascripts/app/controllers/editor.js
+++ b/app/assets/javascripts/app/controllers/editor.js
@@ -739,13 +739,13 @@ angular.module('app')
         return;
       }
       this.loadedTabListener = true;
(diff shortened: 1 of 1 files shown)


Arrow7000:
See this gif for how it works

standard-notes-list

Arrow7000:
See gif for how this functionality behaves

standard-notes-list

PR #263, opened 2018-12-10
fix(desktop): search highlight on pagination

jh-code:
Referencing issue standardnotes/bounties#20

Just triggering it on pagination caused some issues because typing in the search box can trigger it as well so I added a flag to prevent that.

### DIFF
--- a/app/assets/javascripts/app/controllers/notes.js
+++ b/app/assets/javascripts/app/controllers/notes.js
@@ -34,6 +34,7 @@ angular.module('app')
   .controller('NotesCtrl', function (authManager, $timeout, $rootScope, modelManager, storageManager, desktopManager) {
 
     this.panelController = {};
+    this.searchEntered = false;
 
     $rootScope.$on("user-preferences-changed", () => {
       this.loadPreferences();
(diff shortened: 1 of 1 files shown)


mobitar:
What about checking if `this.noteFilter.text` is empty or not on pagination instead of creating a new var `searchEntered`?

jh-code:
Pagination can happen multiple times while typing into the search box, so checking if the filter is empty or not wouldn't work. The issue arises when you enter some text, press enter, and then try to search for a new string. After each character press that triggers pagination, it will unfocus on the input. The flag prevents all of that and only lets it run after someone has search and not entered anything new into the search box.

mobitar:
Oh I see, `searchEntered` means "search has submitted". I thought it meant "user has entered into search", but it means "user has entered out of search". Can we rename that var to something like `searchReturned` or `searchSubmitted`?

jh-code:
I was thinking about renaming that before honestly, I figured that name wouldn't be obvious. Nonetheless, just renamed it to `searchSubmitted` and pushed it.

mobitar:
Great, thanks. Also, have you confirmed whether the ordering of 

```

jh-code:
Good call, I wasn't able to notice any issues, but just to be safe I added that change as well.

mobitar:
Nice work, thanks @jh-code!

PR #272, opened 2019-02-06
Cleanup after bower removing

dajvido:
standardnotes

In commit https://github.com/standardnotes/web/commit/4f423301674cdae960d7e93de9d75728d6e35b07 it was removed only partialy making it fail on execution.


mobitar:
Good catch, thanks!

PR #277, opened 2019-02-20
Fixed some vulnerabilities in packages

gustawdaniel:
Before: found 324 vulnerabilities (5 low, 315 moderate, 4 high)
Now: found 3 low severity vulnerabilities

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,13 +5,13 @@
   "requires": true,
   "dependencies": {
     "JSONStream": {
-      "version": "1.3.3",
-      "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.3.tgz",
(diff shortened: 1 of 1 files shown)


mobitar:
You would have to commit your package.json. package-lock.json is generated based on package.json.

gustawdaniel:
But I do not want to change any version of packages in `package.json`. I only run the command:

```
npm audit fix
```

Notice that all packages are installed based on the content of `package-lock.json`.

mobitar:
I don't think updating dependencies of dependencies is the right idea. You'd have to update the parent, if there is an update available. You can't be sure that the sub-dependency updates don't have adverse affects on the parent.

https://docs.npmjs.com/auditing-package-dependencies-for-security-vulnerabilities

gustawdaniel:
Ok.

PR #298, opened 2019-05-30
fix: clear desktop search highlight when tag is selected

### LINKED ISSUE #271: Searching then clicking tag clears search but leaves highlighting
In the Windows app, a certain search workflow will cause the search field to be cleared but highlighting to remain (until another search is done):

1) Search for any term that returns results
2) See which tags are used among all of the notes in the search results
3) Click on a tag that is used
4) Search field gets cleared, but the highlighting will remain until another search is done.

Seems like expected behavior would be to either not clear the search field or to get rid of the highlighting once another tag is clicked.

Related: https://github.com/standardnotes/desktop/issues/238

dagjaneiro:
Removes the search highlight in the desktop app when a tag is selected

Closes #271

### DIFF
--- a/app/assets/javascripts/app/controllers/notes.js
+++ b/app/assets/javascripts/app/controllers/notes.js
@@ -309,6 +309,7 @@ angular.module('app')
       }
 
       this.noteFilter.text = "";
+      desktopManager.searchText();
 
       this.setNotes(tag.notes);
 


mobitar:
Accepted under https://github.com/standardnotes/web/commit/f84068120d00e09c69cd8944a85d64e0a176a1eb

PR #304, opened 2019-07-31
Added support for theme color on Firefox progress bar

bryvin:
Before/After:

progressbars

Fixes: https://github.com/sn-extensions/simple-task-editor/issues/22


mobitar:
Nice, thanks!

PR #308, opened 2019-09-12
Dock Icon Titles/Names

bryvin:
Adds a title attribute to the dock icons in the footer to make it easier to identify which icon is for which theme. Requested: https://github.com/standardnotes/forum/issues/687

Excuse the added in mouse cursor ( print screen doesn't capture mouse D: )
example


mobitar:
Nice solution! Thought we would have to implement a custom overlay for this.

PR #335, opened 2019-12-04
Add interval to check document focus for autolock

bryvin:
This creates a 1s interval to track the focus of the document. When it loses focus it triggers autolock if required.

I have a few ideas on how to improve this to not have to poll every 1s that will need testing but wanted to get this pushed to make sure we lock where required.


mobitar:
What is the basis of the other ideas?

bryvin:
Couple options that may also work together:

- Use the blur events that proved unreliable due to iframe's taking focus from the main document and only create this new interval for checking focus when the main window had a blur event. The only reason that wasn't a completely viable option is because of the iframe issue.. so this could help if the user is using anything in the native UI or the plain editor it wouldn't need to poll.
- Extending the interval to be the time of the autolock timer and saving the time of the last check for the focus. This would need some more thought and testing.
- Only need to poll for focus if autolock timer is even enabled

PR #342, opened 2019-12-24
actionsMenu: integration for new privilege "use decrypted actions"

### LINKED ISSUE #340: Add "Use Decrypted Actions" to Manageable Privileges
I'm not sure how much security could be gained from this, but it seemed to me that those options like Publish to Blog hanging around in the Actions menu bar is pretty prone to misoperation. It would be nice to be able to at least require extra confirmation before carrying out any action that could lead to catastrophic results (e.g. publishing backed-up account secrets or TOTP tokens by mistake). The ability to set extra authentication requirements before using any action that needs decrypted access would be a nice feature to have integrated by default.

PeterCxy:
Fixes #340.

Dist files are not rebuilt, same as the patch in snjs. Please apply
updates in snjs first before shipping this commit.

### DIFF
--- a/app/assets/javascripts/app/directives/views/actionsMenu.js
+++ b/app/assets/javascripts/app/directives/views/actionsMenu.js
@@ -1,3 +1,4 @@
+import { PrivilegesManager } from '@/services/privilegesManager';
 import template from '%/directives/actions-menu.pug';
 
 export class ActionsMenu {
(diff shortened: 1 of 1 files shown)


baptiste-grob:
We're nearing the end of a big refactor so this can't be merged in anymore unfortunately (It would have to be re-written in TypeScript and target the develop branch). Thank you for your interest in improving SN!

PR #344, opened 2019-12-31
fix: empty tags not being able to create a new dummy note

bryvin:
Related to user reports on Slack regarding not being able to create a note when using a newer account with empty tags.


app/assets/javascripts/app/controllers/notes.js
```diff
@@ -449,10 +449,15 @@ export class NotesPanel {
...
     }
 
-    this.createNewNote = function() {
-      if(this.selectedNote && this.selectedNote.dummy) {
+    this.createNewNote = function(fromTagChange = false) {
+      // If this is called from a tag change, we always want to get through and add a new dummy note if required
+      // otherwise if the user added a new empty tag and switched to that tag's view with a current dummy note
+      // selected, after the change the view would be completely empty (no dummy note). This would also prevent
+      // a dummy note from being created.
+      if(!fromTagChange && this.selectedNote && this.selectedNote.dummy) {
```
mobitar: Would a better solution be to destroy the dummy note on tag change? That would seem more foundationally solid, versus a patch that requires the caller to call the createNewNote function correctly.

bryvin: It does delete the other dummy note and creates a new one. This happens on tag changes because the dummy note automatically received the tag. If it didn't delete everytime you changed tags it would be adding each tag you visit to the dummy note.

The issue currently is that it tries to create a new dummy note while you still have a selectedNote (which can be a dummy note from the previous tag). Because of this the new dummy note that should be created for the new tag isn't. This happens because the if statement that returns when you were "coming" from a dummy note already.

PR #347, opened 2020-01-10
Use overflow-y: auto instead of overflow-y: scroll

MaxLap:
This is for non-webkit browser.
"scroll" means always show the scrollbar, even when not needed. "auto" will only show when it is needed.

Right now, the scrollbar can be very distracting on Firefox.


mobitar:
Can you describe the before and afters, with screenshots/videos if possible? This is not an area I want to touch without extensive testing, due to the rigor involved in getting to the current point.

MaxLap:
Alright, here it goes! All of this is in Firefox.

I made the images to show that the problem also happens with the Folders component. I forgot to do the same thing for the regular tab list, but it's the same thing and the same fix.

This PR fixes the default tag list and the list of nodes, while https://github.com/sn-extensions/folders-component/pull/21 solves it for Folders.

This is before. You can see that the scrollbars are appearing on hover even if there is nothing to scroll through.
Before with free space

This is again before. Now the vertical scroll bars are appearing and they are useful, but the horizontal one is still not needed but appears.
Before without free space

This is after the fix. You can see that no scroll bar appear because they are not needed.
After with free space

This is after the fix. You can see that only the vertical scroll bars appear, since they are needed, and no horizontal one.
After without free space

mobitar:
Nice, these definitely help :) I'll take a look at this and test it out early next week hopefully.

mobitar:
Would you be able to fix conflicts and commit again? Basically all we've done in master is removed the nested `app` folder, and moved everything up one directory.

mobitar:
Actually I've just done it instead here: https://github.com/standardnotes/web/commit/e7cb54f9429c222168271bf0cc658ce6d52b355e

MaxLap:
Thank you. I was trying to do it, but my dist files were getting very different when i tried building the app.

PR #350, opened 2020-01-19
When adding a new Tag make sure originalTagName is null

kenanM:
https://github.com/standardnotes/bounties/issues/49


mobitar:
@kenanM thanks for taking a stab at this. I'll need to look into the best way to handle it. At first glance, the tags controller is meant as a place for managing many tags, so introducing a persistent variable that tracks `originalTagName` is somewhat obscure and out of place.

mobitar:
This issue has been fixed in https://github.com/standardnotes/web/commit/05e5f311ad0a4c14340f657979cf8d6ccf563c15

PR #357, opened 2020-02-04
Lint task

baptiste-grob:
These commits add `eslint --fix` as an npm script so we can do linting regardless of what tool we're using to edit code.
I've also run the script once, and it fixed every stylistic warning it could. I've reviewed every one of them and checked them in by type of change.

### DIFF
--- a/app/assets/javascripts/controllers/notes/note_utils.js
+++ b/app/assets/javascripts/controllers/notes/note_utils.js
@@ -102,7 +102,7 @@ function stringIsUuid(text) {
   const matches = text.match(
     /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
   );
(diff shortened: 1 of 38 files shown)


app/assets/javascripts/controllers/notes/note_utils.js
```diff
@@ -102,7 +102,7 @@ function stringIsUuid(text) {
   const matches = text.match(
     /\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b/
   );
-  return matches ? true : false;
+  return !!matches;
```
mobitar: Hmm, I guess this just converts the result to a bool, if it's not already. But a bit hard to read IMO.

baptiste-grob: I agree. Should we change the eslint rules to allow for it?

PR #360, opened 2020-02-07
Add React

baptiste-grob:
This introduces exactly three React component instances to the project. It's not much, so if something goes wrong it shouldn't impact too much of the app (I don't anticipate anything majorly weird to happen, however my experience with the codebase being low I think it's the right approach.) And if nothing goes wrong, that means we can move forward with React confidently.

### DIFF
--- a/app/assets/javascripts/components/SimpleMenuRow.js
+++ b/app/assets/javascripts/components/SimpleMenuRow.js
@@ -1,0 +1,55 @@
+import React from 'react';
+import { makeAngularComponent } from './utils';
+
+function SimpleMenuRow({
+  action,
+  label,
+  subtitle,
+  desc,
+  circle,
+  circleAlign,
+  faded,
+  disabled,
+  stylekitClass
+}) {
+  return (
+    <div onClick={action} className="sk-menu-panel-row row" title={desc}>
+      <div className="sk-menu-panel-column">
+        <div className="left">
+          {circle && (!circleAlign || circleAlign === 'left') && (
+            <div className="sk-menu-panel-column">
+              <div className={`sk-circle small ${circle}`} />
+            </div>
+          )}
+
+          <div
+            className={`sk-menu-panel-column ${
+              disabled || faded ? 'faded' : ''
+            }`}
+          >
+            <div className={`sk-label ${stylekitClass ? stylekitClass : ''}`}>
+              {label}
+            </div>
+            {subtitle && <div className="sk-sublabel">{subtitle}</div>}
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export const ngSimpleMenuRow = makeAngularComponent(
+  SimpleMenuRow,
+  'simpleMenuRow',
+  {
+    action: '&',
+    label: '=',
+    subtitle: '=',
+    desc: '=',
+    circle: '<',
+    circleAlign: '=',
+    faded: '=',
+    disabled: '=',
+    stylekitClass: '='
+  }
+);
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/components/SimpleMenuRow.js
```diff
@@ -0,0 +1,55 @@
...
+            </div>
+            {subtitle && <div className="sk-sublabel">{subtitle}</div>}
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export const ngSimpleMenuRow = makeAngularComponent(
```
mobitar: Perhaps you've looked into this more than I have by this point: is `ng` reserved for official Angular byproducts? Or is this the right way to say that ngSimpleMenuRow is SimpleMenuRow converted to an Angular component?

app/assets/javascripts/controllers/footer.js
```diff
@@ -252,13 +274,15 @@ class FooterCtrl {
...
     this.syncManager.sync({
       force: true,
       performIntegrityCheck: true
     }).then((response) => {
       this.$timeout(() => {
-        this.isRefreshing = false;
+        // delaying the status update makes the UI look smoother
+        // when syncing finishes fast.
+        this.setState({ isRefreshing: false });
       }, 200);
```
mobitar: Be sure to factor out constants like this into named constants 👍

Also inline comments should use the [JSDocs](https://jsdoc.app/about-getting-started.html) format: `/** */`. Or perhaps  `/* */`, as that appears to be treated differently.

app/assets/javascripts/controllers/notes/notes.js
```diff
@@ -48,6 +48,7 @@ class NotesCtrl extends PureCtrl {
   constructor(
     $timeout,
     $rootScope,
+    $element,
```
mobitar: What's this used for?

app/assets/javascripts/components/SimpleMenuRow.js
baptiste-grob: It's not reserved no. I just thought naming them like this was a nice way to communicate their purpose.

app/assets/javascripts/controllers/footer.js
baptiste-grob: Which constants are you talking about? `200`? Also can you confirm this comment is correct? I didn't add the timeout myself, but since it took me a bit to understand why it was there I thought I'd document it.

app/assets/javascripts/controllers/notes/notes.js
baptiste-grob: Whoop, this is a leftover from my first stab at integrating React components with controllers. Will remove!

mobitar: I highly recommend using an editor with built-in ESLint support ;) It's really a must-have today for writing safe code.

app/assets/javascripts/controllers/footer.js
mobitar: Yeah the 200 constant. And ha, yeah, I guess that's what it's for. Hard to remember now 😳

app/assets/javascripts/components/SimpleMenuRow.js
mobitar: Perhaps `sn` would be a better prefix. `ng` gives the impression this is coming from Angular core. Hard to find good documentation on this but it's somewhat implicit: https://medium.com/@cironunesdev/angularjs-how-to-name-directives-118ac44b81d4

baptiste-grob: I see your point. However using `sn` would make us lose that implicit documentation that this is an angular component. My first idea was adding an `AngularComponent` suffix, like this: `SimpleMenuRowAngularComponent`. It's kinda long but it's explicit!

PR #382, opened 2020-03-09
Web App Manifest

johnny243:
Currently, there's no `manifest.json` file that includes metadata like `name`, `description`, `icons`, etc.

> A user mentioned that the app icon on Chrome OS doesn't seem to be displaying properly, instead it just shows a black box with the letter S on it.

Here's the Standard Notes app without the `manifest.json` file:

image

---

And here's the App with said file:

image

### DIFF
--- a/public/manifest.json
+++ b/public/manifest.json
@@ -1,0 +1,25 @@
+{
+  "app": {
+    "launch": {
+      "urls": [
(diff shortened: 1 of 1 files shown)


mobitar:
Nice 👍

PR #384, opened 2020-03-10
Lowercase emails

baptiste-grob:
When registering, sends a trimmed + lowercase email.

I believe the syncing server's error message should be updated (instead of doing custom logic on the client) to provide a hint when the user fails to log in with an email containing spaces or uppercase letters. If you're of the same mind @mobitar then I'll log an issue on the repo as well


mobitar:
Actually I'm thinking we put this logic directly into SNJS instead, because if we ship this on web but not mobile, then they might have trouble signing in. Tracking this here: https://github.com/standardnotes/snjs/issues/6

PR #385, opened 2020-03-10
Upgrade stylekit dep to 2.0.22

baptiste-grob:


mobitar:
Should this also update the build files as well as package_lock.json?

baptiste-grob:
I don't include the build files with my PRs to avoid some potentially messy merge conflicts. the lockfile however should have been included, yes, let me push it real quick

PR #390, opened 2020-03-11
Add webpack dev server

baptiste-grob:
Using the webpack dev server during development gives us a few things:

- Automatically reload the page when the bundle changes
- Source maps work (🎉)
- Ruby becomes optional for local development

I think that last one is pretty nice for people who would be interested in contributing. The Ruby community is way smaller than the JS one, our app is made with JS, and setting Ruby up on Windows is not super straightforward. This PR makes it so we only use the node/npm ecosystem to work locally, without breaking the production Rails stuff.

Looking forward, this also gives us access to hot module replacement, which combined with a UI library like React makes for a much better developer experience imo.

The one thing I'm not very enthusiastic about is this new `dev-index.html` that I'm having us copy into the `public` folder. That's my approach for the dev server to work without exposing unnecessary files in production, but I'm very open to suggestions on that front.


mobitar:
Should we just move index.html from root to public? Then we wouldn't need the copy approach?

JaspalSuri:
Making the project easier to contribute to could make for a good blog post or tweet.

baptiste-grob:
> Should we just move index.html from root to public? Then we wouldn't need the copy approach?

I believe this would hijack the rails server in production. If I do 
1. `npm run bundle`
2. `bundle exec rails s`
And go to `localhost:3000`, it serves the `index.html` file instead of kicking off the Rails controller logic. That's why I'm having us do the copy in development only.
An alternative would be to name the file something other than `index.html` (like `dev-index.html`) and for development go to `localhost:3000/dev-index.html` instead. Let me know what you think! Notice that all of these approach don't touch the Ruby parts at all, but that could be another possibility

baptiste-grob:
> Making the project easier to contribute to could make for a good blog post or tweet.

Not sure if this is blogpost-worthy but a tweet could be nice 🙂

mobitar:
I think we should go with a dedicated html file in public, possibly just dev.html. The reason is that even if you do the copy approach in dev, at some point you'll forget you did that and Rails functionality and public folder will collide. It might just cause too much confusion.

I think public/dev.html seems to be a nice compromise?

baptiste-grob:
Agreed 👍

mobitar:
Wait, why not just specify the devServer.contentBase as the [root index.html](https://github.com/standardnotes/web/blob/master/index.html) file?

baptiste-grob:
Ah right, should have brought that up actually. I'm not sure what part of the code is supposed to use these variables, but this line here:

```js
window._extensions_manager_location = "public/extensions/extensions-manager/dist/index.html";
```

Has no effect. When we query the components from the model manager, the extensions "room" always points to `/extensions/extensions-manager/dist/index.min.html`. Which breaks the extensions room when we're not serving content from `public`. If we want to use that index file, I have two questions:

1. should SNJS respect the preference, and make the extensions room url point to whatever `window._extensions_manager_location` is set to?
2. What is the `index.html` used for currently?

mobitar:
That variable is used once to create the extensions manager. After it's created, this variable is no longer queried.

https://github.com/standardnotes/web/blob/master/app/assets/javascripts/services/nativeExtManager.js#L37

index.html is currently used to allow bundling the app as a static application without rails (for GitHub releases for example, or as an npm module)

baptiste-grob:
Aah, so it would definitely be better to use that then. I've read the docs again and there's a way to rewrite urls with the dev server, which is what this new commit does for things pointing to `/extensions`. I think with that set up should be good 👍

mobitar:
Nice 👍Only the readme now I think needs to be updated (to remove dev.html mention)

baptiste-grob:
Oof, right. Sorry about that.

PR #392, opened 2020-03-18
Fix component actions still running after destruction

baptiste-grob:
Fixes [desktop#498](https://github.com/standardnotes/desktop/issues/498)
Basically, `reloadComponent()` being async, it can sometime still be running after `componentView` has been destroyed which can cause some weird (and hard to trace down) problems, our app freeze on desktop being one of them.
I chose to fix this by adding and checking against a boolean which is bad for maintenance but is the least disruptive thing to do code-wise. @mobitar you have a better visibility on how `snjs`'s refactor would/will impact this file than me so if you think we can go further with some kind of cancelation mechanism (which standard JS promises don't have) let me know 👍


mobitar:
Seems like a good solution to me 👍

PR #398, opened 2020-05-05
Webpack configuration

johnny243:


webpack.dev.js
```diff
@@ -0,0 +1,28 @@
...
+      '/extensions': {
+        target: 'http://localhost:3001',
+        pathRewrite: { '^/extensions': '/public/extensions' }
+      }
+    },
+    port: 3001
+  },
+  plugins: [
+    ...config.plugins,
+    new WriteFilePlugin()
```
baptiste-grob, line 19: Why would we need to write the bundle to disk during development?

webpack.dev.js
```diff
@@ -0,0 +1,28 @@
...
+  ],
+  // Watches for changes on snjs. Useful when developing snjs locally.
+  watchOptions: {
+    ignored: [
```
baptiste-grob, line 23: Wait so is it watching or ignoring snjs? Is the objective to rebundle when snjs changes?

webpack.dev.js
```diff
@@ -0,0 +1,28 @@
...
+    proxy: {
+      '/extensions': {
+        target: 'http://localhost:3001',
+        pathRewrite: { '^/extensions': '/public/extensions' }
+      }
+    },
+    port: 3001
+  },
+  plugins: [
+    ...config.plugins,
```
baptiste-grob, line 18: I believe using webpack-merge means you don't have to array-spread your existing config

package.json
```diff
@@ -7,12 +7,11 @@
     "url": "https://github.com/standardnotes/web"
   },
   "scripts": {
-    "start": "webpack-dev-server --progress",
-    "watch": "webpack --mode development -w",
-    "bundle": "webpack --mode production",
+    "start": "webpack-dev-server --progress --open --config webpack.dev.js",
```
baptiste-grob, line 10: Do you like `--open`? I find that as soon as I want to relaunch the server or want to test something out in another browser it gets in the way.

webpack.dev.js
johnny243: > Wait so is it watching or ignoring snjs?

It's ignoring everything except `snjs`

> Is the objective to rebundle when snjs changes?

Yup.

webpack.dev.js
johnny243: You're right :+1:

package.json
johnny243: > Do you like `--open`?

Yeah, but I don't mind opening the browser tab manually either. I'll remove it :+1:

webpack.dev.js
johnny243: Because `webpack` is watching changes that occur in the file system, and not changes in memory. In this case, whenever a new change to the `snjs` module is made, the bundle is update. Immediately after, the bundle in the `web` app is updated too.

The same applies when any change is made within the `web` app. Although, it will not update `snjs` module if no new changes are detected.

PR #401, opened 2020-05-08
autolock: add 20 minute option

mrankine:
I use the autolock feature on Standard Notes desktop. I find 5 minutes is too short (hate retyping my password after a quick browsing break), but an hour is too long (someone could pinch my laptop and access my notes in that time). So here's a PR for a 20 minute option. Tested, seems to work.

Tbh I've mainly done this to learn about the GitHub workflow and contributing to open-source projects #firsttimer.


baptiste-grob:
Hey @mrankine, congrats on making your first pull request :)
To be honest if we were to add another autolock option it would probably be an input field for the user to specify their preferred duration. Because after all, if 20 minutes is valid, why not 25, or 35 as well? It's kind of a never-ending chase 😅
The other thing is that this repo is going under some deep restructuring right now so I'm afraid your changes would get wiped accidentally as well.
I'll be closing this PR, but good job anyway!

mrankine:
No worries @baptiste-grob, I had the same thought about user-specified duration.

PR #409, opened 2020-06-05
refactor: import crypto from sncrypto (dep of snjs)

mobitar:
See https://github.com/standardnotes/snjs/pull/15#issuecomment-639563964


app/assets/javascripts/ui_models/application.ts
```diff
@@ -26,6 +26,7 @@ import {
   KeyboardManager
 } from '@/services';
 import { AppState } from '@/ui_models/app_state';
+import { SNWebCrypto } from 'sncrypto/dist/sncrypto-web';
```
baptiste-grob, line 29: How does this work? `sncrypto` is not in the web project's dependencies.

mobitar: sncrypto is listed as a dependency of snjs, so it should be included in node_modules when snjs is installed.

baptiste-grob: For security and project stability reasons webpack doesn't resolve modules that aren't direct dependencies. What we should do is add `sncrypto` to the web project's dependencies, and if we're worried about api mismatch (unlikely since we're using TypeScript), have snjs declare sncrypto as a peer dependency.

baptiste-grob: Here's the output when I run `npm install && npm start` on this branch:
```
ERROR in ./app/assets/javascripts/ui_models/application.ts
Module not found: Error: Can't resolve 'sncrypto/dist/sncrypto-web' in 'C:\Data\dev\sn\web\app\assets\javascripts\ui_models'
```

baptiste-grob:
We'll also need to publish a new sncrypto version to we can reference `sncrypto/dist/sncrypto-web`

mobitar:
We use git commits instead of npm versioning for sncrypto and snjs

mobitar:
https://github.com/standardnotes/snjs/blob/004/package.json#L51

PR #414, opened 2020-06-17
Fix fonts assets

johnny243:


package.json
```diff
@@ -9,7 +9,7 @@
   "scripts": {
     "start": "webpack-dev-server --progress",
     "bundle": "webpack --mode production",
-    "build": "bundle install && npm install && npm run bundle",
+    "build": "bundle install && npm install && bundle exec rails assets:precompile && npm run bundle",
```
karolsojko, line 12: npm install should be npm ci

Can you also adjust the Dockerfile so it doesn’t duplicate on running these commands?

johnny243: Done :+1:

baptiste-grob:
Would this change work with the webpack-dev-server as well?

karolsojko:
> Would this change work with the webpack-dev-server as well?

I've checked out @johnny243 branch and did `npm ci && npm start` and everything is looking good. Webpack serves from dist so that should be ok.

PR #418, opened 2020-06-23
fix: update CSP to allow action bar to download notes

ningsongshen:
This is regarding an issue pointed out by @not-a-robot-yet and fixes standardnotes/action-bar#8.

Since the Action Bar extension works in the desktop app, I decided to look at what was different in the web app. This was the error I got when using the Action Bar:
```
Content Security Policy: The page’s settings blocked the loading of a resource at data:text/plain;charset=utf-8,%3Cp%20dir… (“frame-src”).
```
I think it can be fixed by allowing `'data:'` in the `frame-src` CSP because this is how the Action Bar saves the note: 
```javascript
pom.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(text));
``` 

I'm not too familiar with the web app, but perhaps you guys could weigh in on the security implications.

### DIFF
--- a/app/assets/templates/directives/component-view.pug
+++ b/app/assets/templates/directives/component-view.pug
@@ -89,7 +89,7 @@ iframe(
   ng-attr-id='component-iframe-{{ctrl.component.uuid}}', 
   ng-if='ctrl.component && ctrl.componentValid', 
   ng-src='{{ctrl.getUrl() | trusted}}', 
-  sandbox='allow-scripts allow-top-navigation-by-user-activation allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-modals allow-forms'
+  sandbox='allow-scripts allow-top-navigation-by-user-activation allow-popups allow-popups-to-escape-sandbox allow-same-origin allow-modals allow-forms allow-downloads'
   )
   | Loading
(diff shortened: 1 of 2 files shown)


mobitar:
The error I get with Chrome is:

> Download is disallowed. The frame initiating or instantiating the download is sandboxed, but the flag ‘allow-downloads’ is not set. See https://www.chromestatus.com/feature/5706745674465280 for more details.

And the solution would be implemented [here](https://github.com/standardnotes/web/blob/develop/app/assets/templates/directives/component-view.pug#L92).

Were you able to test your changes in to see if they worked? Curious why we're not seeing the same error.

ningsongshen:
I was able to test my changes, but I'm using Firefox on Linux so I think that might be where the difference is. I can try adding that and seeing if it works on Firefox and Chrome (to be installed).

ningsongshen:
OK now it works in both Firefox and Chrome. But it seems like both are needed, because I tried it without the CSP change in Firefox and the CSP error showed up again.

baptiste-grob:
(Can't really comment on this one because I don't have a clear idea of what this change implies security-wise so I'm leaving it all to you @mobitar)

PR #419, opened 2020-06-24
fix: initialize userPreferences in preferencesManager

johnny243:
Newly created accounts do not have a `UserPrefs` item:

image

The present PR aims to fix this issue by calling `this.loadSingleton();` if User preferences are `undefined`.

### DIFF
--- a/app/assets/javascripts/services/preferencesManager.ts
+++ b/app/assets/javascripts/services/preferencesManager.ts
@@ -25,6 +25,9 @@ export class PreferencesManager extends ApplicationService {
   }
 
   streamPreferences() {
+    if (!this.userPreferences) {
+      this.loadSingleton();
+    }
     this.application!.streamItems(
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/services/preferencesManager.ts
```diff
@@ -25,6 +25,9 @@ export class PreferencesManager extends ApplicationService {
   }
 
   streamPreferences() {
+    if (!this.userPreferences) {
```
mobitar, line 28: Hmm..it looks like `this.userPreferences` will always be null at this point. In which case the solution should be:

1. Rename `loadSingleton` to `reloadSingleton` to indicate it can be called more than once.
2. Call `reloadSingleton` in `onAppLaunch` above `streamPreferences`

johnny243: Done :+1:

PR #421, opened 2020-06-25
fix: editor-view global display options

johnny243:
This PR fixes the following bugs:

- [x] Options were not toggled the first time the button was clicked (this bug is present in `master` branch)
- [x] The "Monospace Font" option became inverted if toggled (see https://github.com/standardnotes/snjs/pull/25)
- [x] The "Spellcheck" option removed the note editor from the view

### DIFF
--- a/app/assets/javascripts/views/editor/editor-view.pug
+++ b/app/assets/javascripts/views/editor/editor-view.pug
@@ -132,7 +132,7 @@
                 .sk-menu-panel-header-title Global Display
               menu-row(
                 action="self.selectedMenuItem(true); self.toggleWebPrefKey(self.prefKeyMonospace)" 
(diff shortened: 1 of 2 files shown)


johnny243:
> The "Monospace Font" option became inverted if toggled (see standardnotes/snjs#25)

We can take this further by using `WebPrefKey.EditorMonospaceEnabled` to initialize state (can apply for other preferences like `EditorSpellcheck` and `EditorResizersEnabled`). This way, it will not brake whenever the `enum` value is changed.

mobitar:
Did I have the order of unloading backwards 🤔 If so this comment should be updated as well: https://github.com/standardnotes/web/blob/72de9516b8b740756236d82301f22e681dbec3c7/app/assets/javascripts/views/editor/editor_view.ts#L80

PR #423, opened 2020-06-26
feature: application group improvements

radko93:
This PRs adds cleanup for observers and additionally avoids observers being notified twice with the same result.

### DIFF
--- a/app/assets/javascripts/ui_models/application_group.ts
+++ b/app/assets/javascripts/ui_models/application_group.ts
@@ -11,11 +11,12 @@ import {
   ThemeManager
 } from '@/services';
 import { AppState } from '@/ui_models/app_state';
+import { pull } from 'lodash';
 
 type AppManagerChangeCallback = () => void
 
 export class ApplicationGroup {
-  
+
   $compile: ng.ICompileService
   $rootScope: ng.IRootScopeService
   $timeout: ng.ITimeoutService
@@ -50,8 +51,9 @@ export class ApplicationGroup {
     }
     if (this.applications.length === 0) {
       this.createDefaultApplication();
+    } else {
+      this.notifyObserversOfAppChange();
     }
-    this.notifyObserversOfAppChange();
   }
 
   private createNewApplication() {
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/ui_models/application_group.ts
```diff
@@ -121,6 +123,10 @@ export class ApplicationGroup {
     if (this.application) {
       callback();
     }
+
+    return () => {
+      pull(this.changeObservers, callback)
```
baptiste-grob, line 128: Looking at the lodash source, using `pull` seems overkill when we only need to check for referential equality. I think a combination of `indexOf` and `splice` would be more efficient, maybe as a utility function? (see `utils.ts`)

radko93: Updated.

mobitar: SNJS has slowly been in the process of removing dependence on lodash in favor of utils. So we have [`removeFromArray`](https://github.com/standardnotes/snjs/blob/004/lib/utils.ts#L175) that can be imported from SNJS.

app/assets/javascripts/ui_models/application_group.ts
```diff
@@ -125,7 +124,8 @@ export class ApplicationGroup {
     }
 
     return () => {
-      pull(this.changeObservers, callback)
+      const indexOfObserver = this.changeObservers.indexOf(callback);
+      this.changeObservers.splice(indexOfObserver, 1);
```
mobitar, line 128: Should use `removeFromArray`. We use this 14 times already in the web code. No need to use splice directly.

app/assets/javascripts/ui_models/application_group.ts
radko93: Thanks for the info. I will remove lodash in mobile code completely in 004.

mobitar: We don't have to totally phase out lodash if it's difficult. Just sort of a gradual process is good enough. But in most cases I believe snjs's utils collection offers a good replacement for most of the basic lodash utils.

PR #426, opened 2020-07-07
Fix editor stack

baptiste-grob:
The editor stack should now work as expected: each note has its own state (with components being open/closed), and that state is synced between clients/restored on note change.
One of the things I did to help me fix this is I stopped relying on ComponentGroup to reduce the number of directions the data was moving in the editor view: it used to be from and to the component manager, as well as from and to the ComponentGroup. Now it's component manager -> editor view -> component manager. I'll look into using the ComponentGroup as a way to abstract the plumbing and interacting with the component manager so the editor view only has to talk to the ComponentGroup, but only after I'm done with making sure the editors and tags work well. I'll share my findings with @radko93 as well.

### DIFF
--- a/app/assets/javascripts/views/editor/editor_view.ts
+++ b/app/assets/javascripts/views/editor/editor_view.ts
@@ -15,7 +15,7 @@ import {
   ComponentArea,
   ComponentAction,
   WebPrefKey,
-  ComponentMutator
+  ComponentMutator,
 } from 'snjs';
 import find from 'lodash/find';
 import { isDesktopApplication } from '@/utils';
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/views/editor/editor_view.ts
```diff
@@ -381,38 +383,39 @@ class EditorViewCtrl extends PureViewCtrl implements EditorViewScope {
...
       async (items) => {
-        const components = items as SNComponent[];
-        if (!this.note) {
-          return;
-        }
-        /** Reload componentStack in case new ones were added or removed */
-        await this.reloadComponentStack();
+        if (!this.note) return;
+        this.setState({
+          allStackComponents: sortAlphabetically(
```
mobitar, line 388: Should this just be called `stackComponents` instead of `all`? Because if previously it was called `activeStackComponents` then the new naming would not be `all`, but a subset of all that are active.

PR #428, opened 2020-07-11
feat: namespaced keychain

johnny243:
Related: https://github.com/standardnotes/snjs/issues/13


app/assets/javascripts/interface.ts
```diff
@@ -98,18 +96,18 @@ export class WebDeviceInterface extends DeviceInterface {
   }
 
   async getKeychainValue() {
-    const value = localStorage.getItem(KEYCHAIN_STORAGE_KEY);
+    const value = localStorage.getItem(this.keychainStorageKey);
```
baptiste-grob, line 99: Where is `this.keychainStorageKey` defined? I'm not seeing it in this PR or the one over on the SNJS repo https://github.com/standardnotes/snjs/pull/42/files

mobitar: It's a getter: https://github.com/standardnotes/snjs/pull/42/files#diff-1a107e6229e1a63c36015525d8219e16R73

mobitar:
I suppose it might make sense to think about how multiple application namespaces would look like. We will need to persist `namespaces` in localStorage as a top-level key I think (nonwrapped). Then each new application instance would determine its namespace based on the length of existing namespaces.

Something like:

```javascript
/** Before creating an application */
const namespaces = localStorage.get("namespaces") || [];
const namespace = `application-${namepaces.length}`;

const application = new Application(..., namespace, ...);
```

app/assets/javascripts/interface.ts
baptiste-grob: Ooh I skimmed right over that `get` symbol, thanks.

baptiste-grob:
I suppose we can address this when we get to actually running multiple applications at once. I just thought that was the intent of the feature.

app/assets/javascripts/services/bridge.ts
```diff
@@ -1,24 +1,30 @@
 /** Platform-specific (i-e Electron/browser) behavior is handled by a Bridge object. */
 export interface Bridge {
-  getKeychainValue(): Promise<unknown>;
+  getKeychainValue(): Promise<any>;
```
baptiste-grob, line 3: Why the switch to `any`? `unknown` is correct here because we're fetching something from a third-party resource (OS keychain, localstorage). We should treat that result with caution, and at worst it's just a cast to `string` away from being usable.

app/assets/javascripts/ui_models/application.ts
```diff
@@ -56,11 +56,9 @@ export class WebApplication extends SNApplication {
     $timeout: ng.ITimeoutService,
     scope: ng.IScope,
     onDeinit: (app: WebApplication) => void,
-    bridge: Bridge,
+    bridge: BrowserBridge,
```
baptiste-grob, line 59: The reason we're relying on `Bridge` and not `BrowserBridge` is so different clients (namely Electron) can provide their own platform-specific implementations. If you need another property on `Bridge` you should add it there instead of depending on `BrowserBridge`

app/assets/javascripts/web_device_interface.ts
```diff
@@ -1,18 +1,16 @@
...
 
 export class WebDeviceInterface extends DeviceInterface {
 
   private database: Database
 
   constructor(
-    namespace: string,
     timeout: any,
-    private bridge: Bridge
+    private bridge: BrowserBridge
```
baptiste-grob, line 11: Should keep Bridge here too

app/assets/javascripts/web_device_interface.ts
```diff
@@ -100,15 +98,39 @@ export class WebDeviceInterface extends DeviceInterface {
...
+
+  async setNamespacedKeychainValue(value: any) {
+    let keychain = await this.getRawKeychainValue();
+    if (!keychain) {
+      keychain = {};
+    }
+    localStorage.setItem(this.bridge.keychainStorageKey, JSON.stringify({
+      ...keychain,
+      [this.namespace!.identifier]: value,
+    }));
```
baptiste-grob, line 117: Should be `this.bridge.setKeychainValue({ ...keychain, [this.namespace!.identifier]: value })` instead. On desktop we use the OS keychain instead of localStorage, hence the extraction of the keychain methods to a common `Bridge` interface

app/assets/javascripts/web_device_interface.ts
```diff
@@ -100,15 +98,39 @@ export class WebDeviceInterface extends DeviceInterface {
...
 
-  setKeychainValue(value: any) {
-    return this.bridge.setKeychainValue(value);
+  async clearNamespacedKeychainValue() {
+    const keychain = await this.getRawKeychainValue();
+    if (!keychain) {
+      return;
+    }
+    delete keychain[this.namespace!.identifier];
+    localStorage.setItem(this.bridge.keychainStorageKey, JSON.stringify(keychain));
```
baptiste-grob, line 126: Should be `this.bridge.setKeychainValue(keychain);`

app/assets/javascripts/services/bridge.ts
```diff
@@ -1,24 +1,30 @@
...
   }
 
   async clearKeychainValue(): Promise<void> {
-    localStorage.removeItem(KEYCHAIN_STORAGE_KEY);
+    localStorage.removeItem(this.keychainStorageKey);
+  }
+
+  get keychainStorageKey() {
+    return KEYCHAIN_STORAGE_KEY;
   }
```
baptiste-grob, line 29: The keychain storage key is a bridge-specific implementation detail that shouldn't be exposed to other parts of the code (see other comments for the code that relies on this property)

app/assets/javascripts/services/bridge.ts
```diff
@@ -1,24 +1,30 @@
...
   clearKeychainValue(): Promise<void>;
 }
 
 const KEYCHAIN_STORAGE_KEY = 'keychain';
 
 export class BrowserBridge implements Bridge {
-  async getKeychainValue(): Promise<unknown> {
-    const value = localStorage.getItem(KEYCHAIN_STORAGE_KEY);
+
+  async getKeychainValue(): Promise<any> {
```
baptiste-grob, line 12: should return `Promise<unknown>` here too

baptiste-grob (changes requested):
Please also update the snjs dependency, I'm assuming that's why the type check is failing :)

baptiste-grob (approved):

PR #430, opened 2020-07-14
feat: server history support

johnny243:
For https://github.com/standardnotes/snjs/issues/38

image


mobitar:
Will hold off on reviewing this until any SNJS related changes as that may affect this PR.

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -1,42 +1,59 @@
...
+  async fetchServerHistory() {
+    this.setState({
+      fetchingServerHistory: true
     });
-    this.history = history;
+    this.serverHistory = await this.application.historyManager!.serverHistoryForItem(this.item)
+      .finally(() => {
+        this.setState({
+          fetchingServerHistory: false
+        });
```
baptiste-grob, line 55: If we're using promise functions as control flow statements it'd be more consistent with the rest of the codebase to use built-in language constructs, i-e:

```js
try {
  this.serverHistory = await this.application.historyManager!.serverHistoryForItem(this.item)
} finally {
  this.setState({
    fetchingServerHistory: false
  });
}
```

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -103,16 +122,14 @@ class SessionHistoryMenuCtrl implements SessionHistoryScope {
...
-        undefined,
-        undefined,
-        undefined,
-        run,
-        undefined,
-        true, 
-      );
+        "Are you sure you want to save history to disk? This will decrease general " +
+        "performance, especially as you type. You are advised to disable this feature " +
+        "if you experience any lagging."
```
baptiste-grob, line 127: I think we lost the `danger` button accent in the conversion. It's the `true` argument at the end of the method call. BTW, since we're in the web repo you can use the `confirmDialog` function instead which supports named arguments specifically to avoid this kind of confusion. There should be some examples of it in the codebase already.

app/assets/javascripts/directives/views/revisionPreviewModal.ts
```diff
@@ -110,14 +110,12 @@ class RevisionPreviewModalCtrl implements RevisionPreviewScope {
...
       this.application.alertService!.confirm(
-        "Are you sure you want to replace the current note's contents with what you see in this preview?",
-        undefined,
-        undefined,
-        undefined,
-        run,
-        undefined,
-        true,
-      );
+        "Are you sure you want to replace the current note's contents with what you see in this preview?"
```
baptiste-grob, line 113: We lost the `danger` button accent here as well I think

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -0,0 +1,184 @@
...
+import { ItemSessionHistory } from 'snjs/dist/@types/services/history/session/item_session_history';
+import { RemoteHistoryList, RemoteHistoryListEntry } from 'snjs/dist/@types/services/history/history_manager';
+import { confirmDialog } from '@/services/alertService';
+
+interface HistoryScope {
+  application: WebApplication
+  item: SNItem
+}
+
+class HistoryMenuCtrl extends PureViewCtrl implements HistoryScope {
```
mobitar, line 15: PureViewCtrl allows you to pass a type for strict checking of state: https://github.com/standardnotes/web/blob/004/app/assets/javascripts/views/abstract/pure_view_ctrl.ts#L7

Example here: https://github.com/standardnotes/web/blob/004/app/assets/javascripts/views/editor/editor_view.ts#L99

This will make it so that if you try to setState of a property not explicitly pre-defined, TS will throw an error.

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -0,0 +1,184 @@
...
+
+  async openSessionRevision(revision: ItemHistoryEntry) {
+    this.application.presentRevisionPreviewModal(
+      revision.payload.uuid, 
+      revision.payload.content
+    );
+  }
+
+  async openRemoteRevision(revision: RemoteHistoryListEntry) {
+    const itemUuid = this.item.uuid;
```
mobitar, line 72: Doesn't need to be extracted into a const. Should just use this.item.uuid inline.

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.fetchingRemoteHistory = true;
+    revision = await this.application.historyManager!.fetchRemoteRevision(itemUuid, revision);
```
mobitar, line 74: Reassigning input parameter is..weird. Should create new var.

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -0,0 +1,184 @@
...
+    if (vector === 0) {
+      return 'default';
+    } else if (vector === 1) {
+      return 'success';
+    } else if (vector === -1) {
+      return 'danger';
+    }
+  }
+
+  clearItemHistory() {
```
mobitar, line 93: Functions for session related history should be delineated as such. `clearItemSessionHistory`, etc. Same for remote.

app/assets/javascripts/directives/views/historyMenu.ts
```diff
@@ -69,17 +73,20 @@ class HistoryMenuCtrl extends PureViewCtrl implements HistoryScope {
   }
 
   async openRemoteRevision(revision: RemoteHistoryListEntry) {
-    const itemUuid = this.item.uuid;
     this.fetchingRemoteHistory = true;
-    revision = await this.application.historyManager!.fetchRemoteRevision(itemUuid, revision);
+    const remoteRevision = await this.application.historyManager!.fetchRemoteRevision(this.item.uuid, revision);
     this.fetchingRemoteHistory = false;
+    if (!remoteRevision) {
+      this.application.alertService!.alert("The remote revision could not be loaded. Please try again later.");
```
johnny243, line 80: We should alert the user in case the revision can not be loaded for some reason.

PR #432, opened 2020-07-15
fix: switching to plain editor

radko93:
Make it possible to switch to the plain editor if another editor is default


mobitar:
Can you talk more about why the existing solution doesn't work? It doesn't seem coincidental that `isExplicitlyEnabledForItem` is used. I honestly can't remember the exact reasoning at this point. But if we do merge this in we need more context, so that if we need to revert it at some point, we understand why it was introduced in the first place. Talk about what you saw, and how changing this affected that behavior.

radko93:
I tested this again after reinstalling Editors and I can switch the editor now at all times. Closing.

PR #433, opened 2020-07-15
fix: ionicons not found on webpack dev server

johnny243:
- [x] fonts are loading in dev server
- [x] precompile assets after installing dependencies
- [x] assets are served via `/assets` path

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,9 @@
   "scripts": {
     "start": "webpack-dev-server --progress",
     "bundle": "webpack --mode production",
-    "build": "bundle install && npm ci && bundle exec rails assets:precompile && npm run bundle",
+    "precompile:assets": "bundle exec rails assets:precompile",
+    "postinstall": "npm run precompile:assets",
+    "build": "bundle install && npm ci && npm run precompile:assets && npm run bundle",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -9,7 +9,9 @@
   "scripts": {
     "start": "webpack-dev-server --progress",
     "bundle": "webpack --mode production",
-    "build": "bundle install && npm ci && bundle exec rails assets:precompile && npm run bundle",
+    "precompile:assets": "bundle exec rails assets:precompile",
+    "postinstall": "npm run precompile:assets",
+    "build": "bundle install && npm ci && npm run precompile:assets && npm run bundle",
```
baptiste-grob, line 14: Can we avoid running bundle/rails when building the web app? The reason for this is that we want to move away from using rails at all because we're only using it to serve static files. Theoretically you shouldn't even need the rails server for development at this point, just webpack.

mobitar: I believe the rails assets generation is essential for deployment. The idea with `build` is one command for all your bootstrapping needs. So if not `build`, then maybe we can make another command then runs build + precompile.

baptiste-grob: Okay, in that case let's keep it as it is for now.

PR #437, opened 2020-07-28
fix: Remove from repository files generated during build process

karolsojko:
The removed files are being generated during docker build process


mobitar:
So the only issue with this I believe is that the desktop app installs this as a package, but via its GitHub source: https://github.com/standardnotes/desktop/blob/dev/app/package.json#L14

And it depends on the dist files here. We could have the desktop app compile the web files beforehand, but I don't know, you might achieve less consistency this way.

Thoughts @baptiste-grob?

baptiste-grob:
From a development perspective, I think tightening up the web and desktop pipeline is desirable. Building the web app from the desktop app gives us a lot more room to make better integration decisions.
The one thing that's giving me pause is that from a security perspective I feel like having a "single source of truth" (a versioned build on one repo) does have value. But that may just be a hunch.

mobitar:
We could just use npm packages instead of git packages, but in my experience git packages are much more convenient.

baptiste-grob:
For security you mean? From a developer experience it sounds much worse.
Maybe my security concerns are moot if we just focus on having reproducible builds instead. In which case I'm all for having the desktop app build the web app itself. I feel like there's a lot of good things that can come out of it.

mobitar:
Sounds good, then feel free to merge this when you are ready to build in that process in desktop.

PR #439, opened 2020-07-29
fix: reset isTemplateNote when setting a new note

radko93:
This fixes a bug when an editor was created for a template note but is later used for a regular note. Before the editor was still marked as `isTemplateNote`.

### DIFF
--- a/app/assets/javascripts/ui_models/editor.ts
+++ b/app/assets/javascripts/ui_models/editor.ts
@@ -68,8 +68,7 @@ export class Editor {
         references: []
       }
     );
-    this.isTemplateNote = true;
-    this.setNote(note as SNNote);
+    this.setTemplateNote(note as SNNote);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/ui_models/editor.ts
```diff
@@ -94,11 +93,20 @@ export class Editor {
...
+    this.note = note;
+    if (this._onNoteChange) {
+      this._onNoteChange();
+    }
+  }
+
   /**
    * Sets the editor contents by setting its note.
    */
   public setNote(note: SNNote) {
```
mobitar, line 107: What about just passing a boolean parameter `isTemplate` to this function instead? Because `setTemplateNote` is only called once so doesn't really make sense to be it's own function, and it's duplicating 90% of `setNote`.

app/assets/javascripts/ui_models/editor.ts
```diff
@@ -97,8 +96,9 @@ export class Editor {
   /**
    * Sets the editor contents by setting its note.
    */
-  public setNote(note: SNNote) {
+  public setNote(note: SNNote, isTemplate?: boolean) {
```
mobitar, line 99: Do you think this is better than just `isTemplate = false` in the signature and then instead of `this.isTemplateNote = isTemplate ?? false;` it would just be `this.isTemplateNote = isTemplate;`?

radko93: True, forgot about it.

PR #440, opened 2020-07-29
Revert "fix: Remove from repository files generated during build process"

karolsojko:
Reverts standardnotes/web#437


karolsojko:
Reverted since last deployment broke the displaying of the app on dev /cc @mobitar @baptiste-grob

PR #449, opened 2020-08-19
fix: incremental loading for actions

johnny243:
- [x] actions are displayed as soon as they are resolved
- [x] loading spinner is shown per extension

### DIFF
--- a/app/assets/javascripts/directives/views/actionsMenu.ts
+++ b/app/assets/javascripts/directives/views/actionsMenu.ts
@@ -2,7 +2,7 @@ import { WebApplication } from '@/ui_models/application';
 import { WebDirective } from './../../types';
 import template from '%/directives/actions-menu.pug';
 import { PureViewCtrl } from '@Views/abstract/pure_view_ctrl';
-import { SNItem, Action, SNActionsExtension } from 'snjs/dist/@types';
+import { SNItem, Action, SNActionsExtension, UuidString } from 'snjs/dist/@types';
 import { ActionResponse } from 'snjs/dist/@types/services/actions_service';
 import { ActionsExtensionMutator } from 'snjs/dist/@types/models/app/extension';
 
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/directives/views/actionsMenu.ts
```diff
@@ -56,16 +56,17 @@ class ActionsMenuCtrl extends PureViewCtrl implements ActionsMenuScope {
...
+      const extensionsForItem = await this.application.actionsManager!.loadExtensionInContextOfItem(
         extension,
         this.props.item
       );
+      const extensions = this.state.extensions || [];
+      this.setState({
+        extensions: extensions.concat(extensionsForItem)
+      });
     }));
     this.loadingExtensions = false;
```
mobitar, line 69: I think loadingExtensions should not be global, but per extension: https://github.com/standardnotes/web/blob/d192594aebcda9a00b7f320000e72917d4df8f37/app/assets/templates/directives/actions-menu.pug#L10

app/assets/javascripts/directives/views/actionsMenu.ts
```diff
@@ -32,7 +32,7 @@ class ActionsMenuCtrl extends PureViewCtrl implements ActionsMenuScope {
 
   application!: WebApplication
   item!: SNItem
-  public loadingExtensions: boolean = true
+  loadingExtensions: boolean = true
```
mobitar, line 35: This might be removed per comment below, but I believe this can be simplified to `loadingExtensions = true`

app/assets/javascripts/directives/views/actionsMenu.ts
johnny243: `this.loadingExtensions` is used to show a spinner while extensions are loaded.

mobitar: Yes but every extension should have its own loading spinner since they are loaded and displayed individually. Here you are displaying 1 single loading spinner for all extensions, which doesn't make as much sense.

johnny243:
I've removed the `hidden` property from the `extension` class: https://github.com/standardnotes/snjs/pull/87

app/assets/templates/directives/actions-menu.pug
```diff
@@ -36,5 +33,5 @@
       menu-row(
         faded='true', 
         label="'No Actions Available'", 
-        ng-if='extension.actionsWithContextForItem(self.props.item).length == 0'
+        ng-if='extension.actionsWithContextForItem(self.item).length === 0'
```
mobitar, line 36: Does `===` work in templates? Not actually sure...something to keep in mind if we experience issues here.

baptiste-grob (changes requested):
Please upgrade SNJS and make sure type info is still correct

app/assets/templates/directives/actions-menu.pug
johnny243: Yeah, it works on templates. In this case I think we should be fine with `==`, which checks the value is equal to 0.

baptiste-grob (approved):

PR #454, opened 2020-08-27
feat: reduce number of steps needed to get going for local development

baptiste-grob:
Using a combination of .env files and a webpack html template with sensible defaults makes for the most straightforward development quick-start experience possible

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,19 +26,18 @@
   <meta ng-bind="title" content="Standard Notes" name="application-name" />
 
   <title>Dev · Notes · Standard Notes</title>
-
-  <script>
-    window._default_sync_server = "https://sync.standardnotes.org";
-    window._extensions_manager_location = "public/extensions/extensions-manager/dist/index.html";
-    window._batch_manager_location = "public/extensions/batch-manager/dist/index.min.html";
-  </script>
-
-  <script src="/dist/javascripts/app.js"></script>
-  <link rel="stylesheet" media="all" href="/dist/stylesheets/app.css" debug="false" />
-
 </head>
 
-<body>
+<body
+  data-default-sync-server="<%= env.DEV_DEFAULT_SYNC_SERVER %>"
+  data-extensions-manager-location="<%= env.DEV_EXTENSIONS_MANAGER_LOCATION %>"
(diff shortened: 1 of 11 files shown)


index.html
```diff
@@ -26,19 +26,18 @@
...
+<body
+  data-default-sync-server="<%= env.DEV_DEFAULT_SYNC_SERVER %>"
+  data-extensions-manager-location="<%= env.DEV_EXTENSIONS_MANAGER_LOCATION %>"
+  data-batch-manager-location="<%= env.DEV_BATCH_MANAGER_LOCATION %>"
+>
   <script>
-    window._default_sync_server = "https://sync.standardnotes.org";
-    window._extensions_manager_location = "public/extensions/extensions-manager/dist/index.html";
-    window._batch_manager_location = "public/extensions/batch-manager/dist/index.min.html";
+    window._default_sync_server = document.body.dataset.defaultSyncServer || "https://sync.standardnotes.org";
```
radko93, line 37: This should not be used anymore actually.

baptiste-grob: Thanks, I've added a commit to set this value in SNJS directly. We still need it on web because this can also be set up by rails at runtime (i-e we can't set the value during build time because it would cripple the docker image)

package.json
```diff
@@ -7,11 +7,12 @@
     "url": "https://github.com/standardnotes/web"
   },
   "scripts": {
+    "setup": "npm run submodules && npm install",
     "start": "webpack-dev-server --progress --config webpack.dev.js",
     "watch": "webpack -w --config webpack.dev.js",
     "bundle": "webpack --config webpack.prod.js",
     "build": "bundle install && npm ci && bundle exec rails assets:precompile && npm run bundle",
-    "submodules": "git submodule update --init --force --remote",
+    "submodules": "git submodule update --init",
```
mobitar, line 15: I've found without force remote it struggles to pull changes to submodules sometimes. Might be good to just keep this if it's generally harmless.

baptiste-grob: the problem with --remote is that it actually updates the submodule's HEAD to the repo's latest commit which in many cases is not what we want. This issue came up because the web repo is now a submodule as well and I can't just pull in any develop commit from that one.
I could keep --force in though.

PR #458, opened 2020-09-01
fix the docker-compose steps

### LINKED ISSUE #457: Steps for docker-compose don't work
I followed the steps for running locally with docker-compose, and they don't work. This is the error I see.

```
WARNING: The PORT variable is not set. Defaulting to a blank string.
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.app.ports is invalid: Port ranges don't match in length
```

The `.env` file needs to be created before you run `docker-compose build`.

Just a minor documentation change. I could open a PR.

jeet-parekh:
Fixes #457.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,15 +85,16 @@ Then open your browser to `http://localhost:3001`.
 
 ### Running Locally with Docker
 
-To run the app locally with Docker, first build the image by running:
-```
-docker-compose build
-```
+To run the app locally with Docker, first create a configuration file:
 
-Then configure and start the application by typing:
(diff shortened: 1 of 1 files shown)


jeet-parekh:
Question. Running locally with docker-compose... is it for development? Or for direct use?

karolsojko:
Good catch - thank you for you contribution 👍

karolsojko:
`docker-compose` is for local development purposes only - if you need to run a production version please use our Official Docker Hub images

PR #462, opened 2020-09-04
fix: prevent importing backup files with version > 003 

baptiste-grob:
This code is not future-proof but because `master` is so different from `develop` I'll have to develop the same thing again on `develop` anyway, where I'll actually check the account version (or do something else that doesn't have hardcoded version numbers)

### DIFF
--- a/app/assets/javascripts/directives/views/accountMenu.js
+++ b/app/assets/javascripts/directives/views/accountMenu.js
@@ -20,6 +20,7 @@ import {
   STRING_GENERATING_REGISTER_KEYS,
   StringImportError
 } from '@/strings';
+import { STRING_IMPORT_FAILED_NEWER_BACKUP } from '../../strings';
 
 const ELEMENT_ID_IMPORT_PASSWORD_INPUT = 'import-password-request';
 
@@ -204,7 +205,7 @@ class AccountMenuCtrl extends PureCtrl {
         text: STRING_NON_MATCHING_PASSWORDS
       });
       return;
-    }    
+    }
     await this.setFormDataState({
       confirmPassword: false,
       status: STRING_GENERATING_REGISTER_KEYS,
@@ -336,7 +337,7 @@ class AccountMenuCtrl extends PureCtrl {
   }
 
   /**
-   * @template 
+   * @template
    */
   async importFileSelected(files) {
     const run = async () => {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/directives/views/accountMenu.js
```diff
@@ -377,6 +377,17 @@ class AccountMenuCtrl extends PureCtrl {
   }
 
   async performImport(data, password) {
+    if (data.keyParams) {
+      if (Number(data.keyParams.version) > 3) {
```
mobitar, line 381: You should be able to use this function instead: https://github.com/standardnotes/snjs/blob/master/lib/protocol/manager.js#L25

app/assets/javascripts/directives/views/accountMenu.js
```diff
@@ -377,6 +377,17 @@ class AccountMenuCtrl extends PureCtrl {
   }
 
   async performImport(data, password) {
+    if (data.keyParams) {
```
mobitar, line 380: `keyParams` is the name of `auth_params` as of 004. In <= 003 it's still called `auth_params`.

app/assets/javascripts/directives/views/accountMenu.js
```diff
@@ -377,6 +377,17 @@ class AccountMenuCtrl extends PureCtrl {
...
+      if (Number(data.keyParams.version) > 3) {
+        this.setState({
+          importData: null
+        });
+        this.alertManager.alert({
+          text: "This backup file was created using a later version of the application and cannot be imported."
```
mobitar, line 386: "This backup file was created using a newer version of the application and cannot be imported here. Please update your application and try again."

app/assets/javascripts/strings.js
```diff
@@ -49,3 +49,5 @@ export function StringImportError({errorCount}) {
 
 /** @password_change */
 export const STRING_FAILED_PASSWORD_CHANGE = "There was an error re-encrypting your items. Your password was changed, but not all your items were properly re-encrypted and synced. You should try syncing again. If all else fails, you should restore your notes from backup.";
+
+export const STRING_IMPORT_FAILED_NEWER_BACKUP = "This backup file was created using a newer version of the application and cannot be imported here. Please update your application and try again.";
```
mobitar, line 53: Why not match this with `develop` value to make merging easier?

baptiste-grob: I agree in principle but merging won't be feasible because we're going from a .js to .ts file (and with significant changes), so every change I make in `master` I backport immediately to `develop`. I don't think it's much of an issue here.

mobitar (approved):

app/assets/javascripts/strings.js
mobitar: I betcha when we go to merge whenever from now you're going to wonder, why on earth did I set these string values different if I created them at the same time 🤔

PR #468, opened 2020-09-11
feat: snjs app groups

mobitar:
Extracts app group from web app into SNJS, and uses that API instead.

Allows for multi-account usage. This will be a feature that we will comment out before final release (requires more careful UX planning). It's really useful for testing however, especially when you want to use the beta desktop app, and have it sign into both your production account and dev account. This way we can test 003 and 004 accounts simultaneously on the same beta desktop/web build.

Designing the app for multi-account usage also helped capture some design faults around deinitialization.

Related: https://github.com/standardnotes/snjs/pull/109


baptiste-grob:
Checked out the branch and ran `npm install`.

On an existing install all my content had a 'Missing Keys' error, but it may be because I'm on a dev install.

Footer view crashes on app start:
image

Clicking _Sign out_ Attempting to sign out triggers these errors and changes the sign out button copy to _Clear Session Data_
image

mobitar:
These changes are not backward compatible with previous dev builds. So you'll have to clear all local data and restart app. We can do this because we're still in development. Otherwise we would have had to create migrations.

However should work if you're going straight from 3.4.x -> this build, although testing this flow in general was on my queue.

mobitar:
Oh also need to update SNJS dep. Was using npm link locally. One sec.

mobitar:
Updated. Was using much older SNJS so you should see much different behavior now. Will merge conflicts before merge.

Come to think about it, it may very well be backward compatible with previous dev builds, but just saying it wasn't a particular goal, since it wasn't necessary.

app/assets/javascripts/ui_models/application.ts
```diff
@@ -27,11 +28,12 @@ import {
 import { AppState } from '@/ui_models/app_state';
 import { SNWebCrypto } from 'sncrypto/dist/sncrypto-web';
 import { Bridge } from '@/services/bridge';
+import { DeinitSource } from 'snjs/dist/@types/types';
 
 type WebServices = {
   appState: AppState
   desktopService: DesktopManager
-  lockService: LockManager
+  autolockService: AutolockService
```
baptiste-grob, line 36: Do you prefer the term Service to Manager? :) (not an issue, just curious)

app/assets/javascripts/ui_models/application.ts
```diff
@@ -103,7 +97,7 @@ export class WebApplication extends SNApplication {
     /** Allow our Angular directives to be destroyed and any pending digest cycles
      * to complete before destroying the global application instance and all its services */
     setImmediate(() => {
-      super.deinit();
+      super.deinit(source);
     })
```
baptiste-grob, line 101: BTW setImmediate is [not standard](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate) and if all you want is to run code on the next event cycle you can use setTimeout with no specified delay (defaults to 0)

app/assets/javascripts/ui_models/application.ts
```diff
@@ -257,4 +251,15 @@ export class WebApplication extends SNApplication {
...
     angular.element(document.body).append(el);
   }
+
+  public openAccountSwitcher() {
+    const scope = this.scope!.$new(true) as Partial<AccountSwitcherScope>;
+    scope.application = this;
+    const el = this.$compile!(
+      "<account-switcher application='application' "
+      + "class='sk-modal'></account-switcher>"
+    )(scope as any);
```
baptiste-grob, line 261: Is there no way to create and insert an angular view other than to compile a string at runtime? 😱

app/assets/javascripts/views/abstract/pure_view_ctrl.ts
```diff
@@ -63,9 +63,15 @@ export class PureViewCtrl<P = CtrlProps, S = CtrlState> {
...
     return new Promise((resolve) => {
-      this.stateTimeout = this.$timeout(resolve);
+      this.stateTimeout = this.$timeout(() => {
+        /**
+         * State changes must be *inside* the timeout block for them to be affected in the UI
+         * Otherwise UI controllers will need to use $timeout everywhere
+         */
+        this.state = Object.freeze(Object.assign({}, this.state, state));
+        resolve();
+      });
```
baptiste-grob, line 74: Wait what? Was this an issue you encountered during the development of this feature? Because otherwise changing state outside the timeout block means we don't have to await a whole angular digest cycle when we're just manipulating values.

app/assets/javascripts/views/account_switcher/account_switcher.ts
```diff
@@ -0,0 +1,105 @@
...
+    $element: JQLite,
+    $timeout: ng.ITimeoutService,
+    private mainApplicationGroup: ApplicationGroup
+  ) {
+    super($timeout);
+    this.$element = $element;
+    this.removeAppGroupObserver = mainApplicationGroup.addApplicationChangeObserver(() => {
+      this.activeApplication = mainApplicationGroup.primaryApplication as WebApplication;
+      this.reloadApplications();
+    });
```
baptiste-grob, line 31: Since deinit is called during $onDestroy this should probably be inside $onInit instead

app/assets/javascripts/web_device_interface.ts
```diff
@@ -14,16 +14,23 @@ export class WebDeviceInterface extends DeviceInterface {
...
+  private databaseForIdentifier(identifier: ApplicationIdentifier) {
+    return this.databases.find(database => database.databaseName === identifier)!;
   }
 
   deinit() {
     super.deinit();
-    this.database.deinit();
+    for(const database of this.databases) {
+      database.deinit();
+    }
```
baptiste-grob, line 32: So does this mean DeviceInterface only gets deinited once for _all_ applications? If it's still on an app-by-app basis I assume this would wreck every app instance.

baptiste-grob (approved):
Before merging, please ensure the following scenarios work:

- Switching from a 3.4.x web app w/ account to this version
  - Also upgrade encryption, just in case
- Going from a 3.4.x desktop app to this version

app/assets/javascripts/views/abstract/pure_view_ctrl.ts
baptiste-grob: Actually please address this one before merging. It may break other parts of the app.

johnny243 (commented):
> I know UI/UX will change in the future, but can we make the application name clickable? Currently, you have to click the area around the title in order to switch accounts.

Peek 2020-09-14 10-56

app/assets/javascripts/ui_models/application.ts
mobitar: Oh interesting..

app/assets/javascripts/ui_models/application.ts
mobitar: The only other way I know of is to insert it in a template with "ng-if" and have that ng-if condition change to true when you want to show it. But I think I would prefer this dynamic way more.

app/assets/javascripts/views/abstract/pure_view_ctrl.ts
mobitar: So the history of this function was:
- it first performed the state change in the timeout 
- after I handed it off to you and you got familiar with it you thought it would make sense to put it outside the timeout. I was hands-off at that point and couldn't really test it but it seemed like a positive change
- after building my feature, it took me back to the original design purpose: setState is designed to trigger a UI refresh, the same way it does in React via render(). If you call setState() in Angular from a callback, it won't update the UI, unless you do it in a $timeout block. For example, [this](https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/views/application_group/application_group_view.ts#L20) wouldn't work if not wrapped in $timeout. Or perhaps something simpler like [this](https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/views/footer/footer_view.ts#L113) wouldn't work either. If it does work, it's just timing, since the `checkForSecurityUpdate` is local and will return quickly. But you can test this by creating an async function that takes 1s to return, and doing a state change in a `then`. It won't work unless you put the state change in a $timeout.

So if you want a setState for a purpose other than updating the UI, we should probably create another function for it, `setStateImmediate` or something like that.

mobitar: This was the function I was looking for: https://github.com/standardnotes/web/blob/app-group/app/assets/javascripts/views/account_switcher/account_switcher.ts#L30 Wouldn't work unless state change happen in $timeout block.

app/assets/javascripts/ui_models/application.ts
mobitar: The distinction I decided on after a lot of self-quarreling was:

- a manager manages things for you and runs things without you being aware of them, like registering for an observer and responding to changes
- a service centers more on how other parts of the code can use it to do what they want

Obviously there will be a lot of overlap so it's just a guideline. But I suppose that by these guidelines the autolock service is more of a manager than a service.

app/assets/javascripts/views/abstract/pure_view_ctrl.ts
baptiste-grob: Okay. We'll have to keep an eye out then. This could break other unrelated code.

PR #481, opened 2020-10-01
improved docs

sajal243:


baptiste-grob:
😂 The compliment is much appreciated! I'm afraid we can't take ownership of it however 😄

PR #530, opened 2021-03-04
feat: notes display criteria

mobitar:
Incorporates SNJS's new [NotesSearchCriteria feature](https://github.com/standardnotes/snjs/pull/232). This PR doesn't include the updated SNJS version, since that's not yet published. So if you'd like to test it you'll need to yarn link locally.

@antsgar a similar change will be necessary in mobile.


baptiste-grob (approved):
All of this looks good to me but should not be merged before the related SNJS PR has been merged (which will make the tsc check pass) 👍

PR #535, opened 2021-03-19
fix: reload when firefox pinned tab becomes visible

mobitar:
@baptiste-grob please merge after approve.


baptiste-grob:
Unless I'm missing something, this would reload the tab every time the user switched off of it?

mobitar:
`reloadHiddenFirefoxTab` is called only on application launch. If the application is launched and the document is hidden, a visibility observer is added. Once the document is visible it will reload the page (thus deregistering the observer). Then the application starts again and this time does not enter into the `reloadHiddenFirefoxTab` flow.

baptiste-grob:
Aah ok, gotcha!

baptiste-grob (approved):

PR #540, opened 2021-04-05
feat: search options

baptiste-grob:


app/assets/javascripts/components/SearchOptions.tsx
```diff
@@ -0,0 +1,154 @@
...
+      setOpen(false);
+    }
+  }
+
+  return (
+    <Disclosure
+      open={isOpen}
+      onChange={() => {
+        const { height } = buttonRef.current.getBoundingClientRect();
+        setOptionsPanelTop(height + 4);
```
mobitar, line 109: Magic value `4` 👀 Maybe should be extracted into const?

app/assets/javascripts/components/SearchOptions.tsx
```diff
@@ -0,0 +1,155 @@
...
+  DisclosurePanel,
+} from '@reach/disclosure';
+import React, { HTMLProps } from 'react';
+
+type Props = {
+  appState: AppState;
+  application: WebApplication;
+};
+
+function Switch(
```
mobitar, line 26: Should be extracted into its own file no? Will need to be done when we need switches elsewhere anyway.

app/assets/javascripts/ui_models/app_state.ts
```diff
@@ -148,6 +148,62 @@ class NoAccountWarningState {
   }
 }
 
+class SearchOptions {
```
mobitar, line 151: So many classes in this file. What's your take on this?

app/assets/javascripts/components/SearchOptions.tsx
baptiste-grob: True. This was originally supposed to live in StyleKit but the React components export is weirdly broken so I postponed the extraction…

app/assets/javascripts/ui_models/app_state.ts
baptiste-grob: I could go either way honestly. 
These classes are only meant to be consumed by AppState so they only exist to separate behavior in a sensible way. In that sense they're not different from one big class with many methods, it's just one holding class containing smaller objects. Extracting each class to its own separate file would the overall intent (to have one global state object) less clear I think.

baptiste-grob: I don't believe extracting each class into its separate file would cause much harm either so I have no quarrel with doing so if you have a stronger opinion about this than I do 🙂

mobitar: Just cleaner to do so imo. Can be in a folder within a app_state dir. Similar to creating a component that has many components: I think it's cleaner if each component were its own file. Especially as the sub components begin to grow. In my experience it's also frustrating to open something like app_state and expect to see AppState but then having to scroll to who knows where to actually begin seeing the AppState interface.

baptiste-grob: Gotcha 👍

mobitar (approved):

app/assets/javascripts/components/SearchOptions.tsx
```diff
@@ -0,0 +1,112 @@
...
+  async function toggleIncludeProtectedContents() {
+    setTogglingIncludeProtectedContents(true);
+    try {
+      await searchOptions.toggleIncludeProtectedContents();
+    } finally {
+      setTogglingIncludeProtectedContents(false);
+    }
+  }
+
+  const [isOpen, setOpen] = useState(false);
```
daianaszwimer, line 47: the name of the hook function should be the `set` + `name of the variable` => `setIsOpen` @baptiste-grob

app/assets/javascripts/components/SearchOptions.tsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  return (
+    <Disclosure
+      open={isOpen}
+      onChange={() => {
+        const { height } = buttonRef.current.getBoundingClientRect();
+        const extraVerticalBreathingRoom = 4;
+        setOptionsPanelTop(height + extraVerticalBreathingRoom);
+        setOpen(!isOpen);
```
daianaszwimer, line 68: @baptiste-grob I don't think we will have issues with this, but just in case, when you send the previous value as a parameter, a good practice is to send a function as a parameter `setOpen((prevIsOpen) => !prevIsOpen);`
https://reactjs.org/docs/hooks-reference.html#functional-updates

app/assets/javascripts/components/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+export type SwitchProps = HTMLProps<HTMLInputElement> & {
+  checked?: boolean;
+  onChange: (checked: boolean) => void;
+  children: ComponentChildren;
+};
+
+export const Switch: FunctionalComponent<SwitchProps> = (
+  props: SwitchProps
+) => {
+  const [checkedState, setChecked] = useState(props.checked || false);
```
daianaszwimer, line 20: Maybe we can set `props.checked`'s default value here, what do you think about that approach? @baptiste-grob 
```
Switch.defaultProps = {
  checked: false,
}
```

app/assets/javascripts/components/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const checked = props.checked ?? checkedState;
```
daianaszwimer, line 21: If we implement this https://github.com/standardnotes/web/pull/540/files#r607887769 we won't need this, right? @baptiste-grob

app/assets/javascripts/components/SearchOptions.tsx
baptiste-grob: Conventionally when it's a boolean setter we omit the `is`. But maybe that's old hat?

app/assets/javascripts/components/Switch.tsx
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    <label className="sn-component flex justify-between items-center cursor-pointer hover:bg-contrast py-2 px-3">
+      {props.children}
+      <CustomCheckboxContainer
+        checked={props.checked != null ? props.checked : checked}
```
daianaszwimer, line 26: This seems a little redundant I think. Aren't you doing the same check on line 21? @baptiste-grob

app/assets/javascripts/components/Switch.tsx
baptiste-grob: I dislike using defaultProps because they break the functional paradigm. Does that even work on functional components actually?

daianaszwimer: Also, same thing that I mentioned on another comment, the name of the function should be `setCheckedState` @baptiste-grob  😄

app/assets/javascripts/components/SearchOptions.tsx
daianaszwimer: From what I see on React's documentation page, there's an example with this same case scenario and they don't omit the `is`
https://reactjs.org/docs/hooks-overview.html#effect-hook => they are declaring ` const [isOnline, setIsOnline] = useState(null);`

app/assets/javascripts/components/Switch.tsx
baptiste-grob: While we don't need this for SearchOptions, it's conventional in component libraries to have components with autonomous behavior by default :)

app/assets/javascripts/components/Switch.tsx
baptiste-grob: You're right, that's a refactoring leftover 😄

app/assets/javascripts/components/Switch.tsx
baptiste-grob: This also doesn't account for using Switch as an uncontrolled component, see my comment below

app/assets/javascripts/ui_models/app_state/sync_state.ts
```diff
@@ -0,0 +1,36 @@
+import { SyncOpStatus } from "@standardnotes/snjs";
+import { action, makeObservable, observable } from "mobx";
+
+export class SyncState {
+  inProgress = false;
+  errorMessage?: string = undefined;
```
daianaszwimer, line 6: `undefined` is used when there's an unintentional "mistake". If you want nullable values you should use `null` or not assign them anything (which will result in having `undefined` as a value) @baptiste-grob

app/assets/javascripts/ui_models/app_state/sync_state.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.errorMessage = status.error?.message;
+    this.inProgress = status.syncInProgress;
+    const stats = status.getStats();
+    const completionPercentage =
+      stats.uploadCompletionCount === 0
+        ? 0
+        : stats.uploadCompletionCount / stats.uploadTotalCount;
+
+    if (completionPercentage === 0) {
+      this.humanReadablePercentage = undefined;
```
daianaszwimer, line 28: Same here ;) https://github.com/standardnotes/web/pull/540/files#r607906141

app/assets/javascripts/ui_models/app_state/sync_state.ts
```diff
@@ -0,0 +1,36 @@
...
+    } else {
+      this.humanReadablePercentage = completionPercentage.toLocaleString(
+        undefined,
```
daianaszwimer, line 31: and here 😅 https://github.com/standardnotes/web/pull/540/files#r607906141

app/assets/javascripts/ui_models/app_state/sync_state.ts
baptiste-grob: Most codebases have to at some point decide if they want to use `null` or `undefined` to signify "the absence of". In our case we chose `undefined` a while ago (most notably in SNJS) so this is just following that convention :)

app/assets/javascripts/components/Switch.tsx
daianaszwimer: > Does that even work on functional components actually

Yes! :)

PR #574, opened 2021-06-07
feat: remove batch manager

antsgar:
- Removed batch-manager submodule
- Removed batched manager related code

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -9,7 +9,6 @@ SECRET_KEY_BASE=test
 APP_HOST=http://localhost:3001
 
 EXTENSIONS_MANAGER_LOCATION=extensions/extensions-manager/dist/index.html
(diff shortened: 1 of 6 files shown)


antsgar:
Figure this should be 3.0.0 since there is some code removal that would be a breaking change?

mobitar (approved):

jmason888:
Apparently, I need to monitor these "features" more closely so that I am not surprised by the removal of a feature that I (occasionally) rely upon (but when I do rely upon it, it is critical).

I now need to see if there is any way for me to restore this functionality, in particular the Backups Explorer, without hosting an entire fork of StandardNotes and related infrastructure, and without waiting for "new features" to be added "in the future".

With all this "simplification" removing features, is StandardNotes in danger of becoming power-user-unfriendly, just like so many products from Big Tech?

PR #575, opened 2021-06-07
fix: don't wait until regaining focus to lock app

antsgar:
- Poll every 10ms (does this sound like a reasonable interval?) to check if the current date is larger than the lock app date, and thus we need to lock the app. Before we used to poll every second, and if the app was visible we'd lock it. By reducing the polling interval we can avoid the app flashing before locking itself, and by checking if we need to lock even if the app isn't visible we're allowing the app to lock itself even when it is on the background
- Remove listeners for window blur and window focus and remove setTimer for locking app, since by reducing the polling interval I think we could rely on polling only

Not sure if this approach makes sense so as usual feedback is very welcome 🙂


antsgar:
@mobitar can I merge this one? 🙂

mobitar (approved):

PR #576, opened 2021-06-11
refactor: split "Account Menu" component into multiple components

vardan-arm:
- refactor: remove `url` and keep only `server`, as they are duplicating each other
- style: change link to button for "Advanced Options"
- style: show cursor as pointer on checkbox and its label
- chore: configure eslint for react hooks


app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+    } else {
+      setIsAuthenticating(false);
+      closeAccountMenu();
+    }
+  };
+
+  const handleAuthFormSubmit = (event:
+                                  TargetedEvent<HTMLFormElement> |
+                                  TargetedMouseEvent<HTMLButtonElement> |
+                                  TargetedKeyboardEvent<HTMLButtonElement>
```
antsgar, line 161: Might be better to use FormEvent

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
+import { confirmDialog } from '@Services/alertService';
+import {
+  STRING_ACCOUNT_MENU_UNCHECK_MERGE,
+  STRING_GENERATING_LOGIN_KEYS,
+  STRING_GENERATING_REGISTER_KEYS,
+  STRING_NON_MATCHING_PASSWORDS
+} from '@/strings';
+import { JSXInternal } from '@node_modules/preact/src/jsx';
+import TargetedEvent = JSXInternal.TargetedEvent;
+import TargetedKeyboardEvent = JSXInternal.TargetedKeyboardEvent;
```
antsgar, line 10: We should probably use FormEvent, ChangeEvent, MouseEvent, KeyboardEvent, etc instead of importing from JSXInternal

app/assets/javascripts/components/AccountMenu/Encryption.tsx
```diff
@@ -0,0 +1,36 @@
+import { FC } from 'react';
+
+type Props = {
+  isEncryptionEnabled: boolean;
+  notesAndTagsCount: number;
+  encryptionStatusString: string | undefined;
```
antsgar, line 6:
```suggestion
  encryptionStatusString?: string;
```

app/assets/javascripts/components/AccountMenu/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+                application={application}
+                setEncryptionStatusString={setEncryptionStatusString}
+                setIsEncryptionEnabled={setIsEncryptionEnabled}
+                setIsBackupEncrypted={setIsBackupEncrypted}
+              />
+              <DataBackup
+                application={application}
+                isBackupEncrypted={isBackupEncrypted}
+                isEncryptionEnabled={isEncryptionEnabled}
+                setIsBackupEncrypted={setIsBackupEncrypted}
```
antsgar, line 140: To keep it cleaner we could move state variables that are shared between several components, such as isBackupEncrypted, isEncryptionEnabled, and their setters, to the shared appState using Mobx

app/assets/javascripts/components/AccountMenu/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+              )}
+              <Encryption
+                isEncryptionEnabled={isEncryptionEnabled}
+                notesAndTagsCount={notesAndTagsCount}
+                encryptionStatusString={encryptionStatusString}
+              />
+              {hasProtections && (
+                <Protections
+                  application={application}
+                  protectionsDisabledUntil={protectionsDisabledUntil}
```
antsgar, line 127: If protectionsDisabledUntil is only used in the Protections component, we could move the related logic (such as the getProtectionsDisabledUntil method, the state variables and the related event observer) to that child component

app/assets/javascripts/components/AccountMenu/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+          <div className="sk-panel-header-title">Account</div>
+          <a className="sk-a info close-button" onClick={closeAccountMenu}>Close</a>
+        </div>
+        <div className="sk-panel-content">
+          <Authentication
+            application={application}
+            server={server}
+            setServer={setServer}
+            closeAccountMenu={closeAccountMenu}
+            notesAndTagsCount={notesAndTagsCount}
```
antsgar, line 101: Both of these deal with AppState variables so I think there's no need to pass them down as props, we can just pass appState and read them from state on the child component

app/assets/javascripts/components/AccountMenu/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+        </div>
+        <ConfirmSignoutContainer application={application} appState={appState} />
+        <Footer
+          appState={appState}
+          application={application}
+          showLogin={showLogin}
+          setShowLogin={setShowLogin}
+          showRegister={showRegister}
+          setShowRegister={setShowRegister}
+          user={user}
```
antsgar, line 154: Same comment probably applies here, if there are shared variables between Authentication, Footer and the parent component we could move them to appState

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  />
+                  <p className="sk-p">Decrypted</p>
+                </label>
+              </div>
+            </form>
+          )}
+          <div className="sk-panel-row" />
+          <div className="flex">
+            <button className="sn-button small info" onClick={downloadDataArchive}>Download Backup</button>
+            <label className="sn-button small flex items-center info ml-2">
```
antsgar, line 135:
```suggestion
            <span type="button" className="sn-button small flex items-center info ml-2">
```
This would be an accessibility improvement 🙂

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
```diff
@@ -0,0 +1,157 @@
...
+              <input
+                type="file"
+                onChange={importFileSelected}
+                style={{ display: 'none' }}
```
antsgar, line 139: Instead of using inline styles, we can add a "hidden" class to _sn.scss

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
```diff
@@ -0,0 +1,157 @@
...
+                  <p className="sk-p">Decrypted</p>
+                </label>
+              </div>
+            </form>
+          )}
+          <div className="sk-panel-row" />
+          <div className="flex">
+            <button className="sn-button small info" onClick={downloadDataArchive}>Download Backup</button>
+            <label className="sn-button small flex items-center info ml-2">
+              <input
```
antsgar, line 136: Accessibility could be improved here too by adding tabindex="0" to make the button keyboard accessible

antsgar:
Nice work @vardan-arm! All in all I think a good next step would be to refactor some of the shared logic between child components by moving it to app state

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+import { WebApplication } from '@/ui_models/application';
+import { StateUpdater, useEffect, useRef, useState } from 'preact/hooks';
+import TargetedMouseEvent = JSXInternal.TargetedMouseEvent;
+import { User } from '@node_modules/@standardnotes/snjs/dist/@types/services/api/responses';
```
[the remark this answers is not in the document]
vardan-arm: Changed in all applicable files. 
Ok, so in my editor `@node_modules/` comes at the beginning of the module import every time after auto-importing it. Will dig into configs later on to not have such imports in the future.

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+import { FC } from 'react';
```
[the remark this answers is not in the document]
vardan-arm: Noted, done

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+                     onChange={handlePasswordChange}
+                     placeholder="Password"
+                     required
+                     onKeyPress={handleKeyPressKeyDown}
+                     onKeyDown={handleKeyPressKeyDown}
+                     ref={passwordInputRef}
+              />
+              {showRegister &&
+              <input className="sk-input contrast"
+                     name="password_conf"
```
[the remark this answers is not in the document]
vardan-arm: Done, if you meant moving `className` to the next line. Otherwise, please clarify :)

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+  useEffect(() => {
+    if (!showLogin && !showRegister) {
+      setPassword('');
+      setPasswordConfirmation('');
+    }
+  }, [showLogin, showRegister]);
+
+  const handleHostInputChange = (event: TargetedEvent<HTMLInputElement>) => {
+    const { value } = event.target as HTMLInputElement;
+    // setUrl(value);
```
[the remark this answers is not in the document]
vardan-arm: Sure, missed this one, thanks!

app/assets/javascripts/components/AccountMenu/Encryption.tsx
vardan-arm: Although in this case there won't be a difference, but having `string | undefined` enforces us to pass `encryptionStatusString` to `Encryption` component, while with `encryptionStatusString?` we can skip that prop and TS won't complain. The point is we expect `encryptionStatusString` to be passed, it just can be either `string` or `undefined`.

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
vardan-arm: When I change `label` to `span`, it simply stops reacting the click event. I researched and in those cases we can append `ref` to the hidden input and on clicking on span, manually triggering `click` event on that input with the `ref`. 
If accessibility improvement is worth adding that logic, I will do that :)

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
vardan-arm: Done

app/assets/javascripts/components/AccountMenu/Encryption.tsx
antsgar: That's a really good point! 🙂

app/assets/javascripts/components/AccountMenu/Authentication.tsx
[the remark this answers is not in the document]
antsgar: Ah I meant, I think there should only be 4 extra spaces for the props?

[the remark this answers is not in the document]
vardan-arm: I ran prettier on my side, should be ok now :)

app/assets/javascripts/components/AccountMenu/index.tsx
vardan-arm: Good point, done!

app/assets/javascripts/components/AccountMenu/index.tsx
vardan-arm: Done.

app/assets/javascripts/components/AccountMenu/index.tsx
vardan-arm: Correct, done

app/assets/javascripts/components/AccountMenu/index.tsx
vardan-arm: Sure, done. Regarding `user` - I decided to move it from `account_menu_state`, as we already can get it by `application.getUser()`. All the others are moved to `account_menu_state`.

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+                     type="password"
+                     placeholder="Confirm Password" required
+                     onKeyPress={handleKeyPressKeyDown}
+                     onKeyDown={handleKeyPressKeyDown}
+                     value={passwordConfirmation}
+                     onChange={handlePasswordConfirmationChange}
+                     ref={passwordConfirmationInputRef}
+              />}
+              <div className="sk-panel-row" />
+              <button className="sn-button small info" onClick={() => {
```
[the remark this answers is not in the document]
vardan-arm: Done

app/assets/javascripts/components/AccountMenu/Authentication.tsx
```diff
@@ -0,0 +1,382 @@
...
+
+  // Reset password and confirmation fields when hiding the form
+  useEffect(() => {
+    if (!showLogin && !showRegister) {
+      setPassword('');
+      setPasswordConfirmation('');
+    }
+  }, [showLogin, showRegister]);
+
+  const handleHostInputChange = (event: TargetedEvent<HTMLInputElement>) => {
```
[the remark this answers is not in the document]
vardan-arm: Using that type was my first approach actually :) But it doesn't work, throws the following error:

```TS2322: Type '(event: ChangeEvent<HTMLInputElement>) => void' is not assignable to type 'GenericEventHandler<HTMLInputElement>'.
   Types of parameters 'event' and 'event' are incompatible.
     Type 'TargetedEvent<HTMLInputElement, Event>' is missing the following properties from type 'ChangeEvent<HTMLInputElement>': nativeEvent, isDefaultPrevented, isPropagationStopped, persist```

The approach you suggested works in case React is used. But here `GenericEventHandler<HTMLInputElement>` comes from `preact` (we can see it on hovering input's `onChange` in code editor), thus we get that difference.

File: app/assets/javascripts/components/AccountMenu/DataBackup.tsx (hunk 10 above)
Comment (vardan-arm) [reply app/assets/javascripts/components/AccountMenu/DataBackup.tsx:134]: Done

File: app/assets/javascripts/components/AccountMenu/DataBackup.tsx (hunk 8 above)
Comment (vardan-arm) [reply app/assets/javascripts/components/AccountMenu/DataBackup.tsx]: So I changed it to `button` to make the styling similar to other buttons and added click and key press handling to import backup.
Please let me know if we need to have `span` instead of button because of some reason :)

File: app/assets/javascripts/components/AccountMenu/Authentication.tsx (hunk 1 above)
Comment (vardan-arm) [reply app/assets/javascripts/components/AccountMenu/Authentication.tsx:156]: Please see the previous comment

File: app/assets/javascripts/components/AccountMenu/Authentication.tsx (hunk 2 above)
Comment (vardan-arm) [reply app/assets/javascripts/components/AccountMenu/Authentication.tsx:10]: I fully agree, but TS complains when using them. Please see my above comments for more details

**Comment by vardan-arm** [comment]
Thanks for your notes, @antsgar ! I've done the required changes, please have a look again :)

File: app/assets/javascripts/ui_models/app_state/account_menu_state.ts [hunk 17]
```diff
@@ -1,30 +1,60 @@
...
 
       notesAndTagsCount: computed
     });
 
-    appEventListeners.push(
+    this.addAppEventObserver();
+    this.streamNotesAndTags();
+  }
+
+  addAppEventObserver = (): void => {
```
antsgar, line 46: Maybe could be called addAppLaunchedEventObserver

app/assets/stylesheets/_sn.scss
```diff
@@ -1,322 +1,17 @@
...
-.w-80 {
-  width: 20rem;
-}
-
-.w-70 {
-  width: 17.5rem;
+.hover-text-underline {
+  &:hover {
+    text-decoration: underline;
+  }
```
antsgar, line 14:
```suggestion
.hover\:underline:hover {
  text-decoration: underline;
}
```

To follow the naming conventions we currently use on StyleKit 🙂

app/assets/javascripts/components/AccountMenu/DataBackup.tsx
antsgar: A button is better actually, nicely done!

app/assets/javascripts/ui_models/app_state/account_menu_state.ts
vardan-arm: Done :)

app/assets/stylesheets/_sn.scss
vardan-arm: Noted, thanks!

antsgar (approved):

PR #588, opened 2021-06-30
feat: implement preferences pane

gorjan5sk:
- Implemented a full screen preferences pane (currently hidden)
- Implemented preferences menu (vertical tab bar)
- Data is provided from a mock state provider as an intermediate solution


app/assets/stylesheets/_preferences.scss
```diff
@@ -0,0 +1,82 @@
+.preferences-view {
```
antsgar, line 1: Styling that will likely only be used on one component shouldn't be extracted on a separate class, but we should rather keep the utility classes in the markup itself. For example, if the styling for .preferences-view is only used in that component, then we wouldn't have a separate preferences-view class, we'd just use all the utility classes in the JSX markup. If there's styling that's likely going to be repeated in several components, then we can extract them to its own class. Here's a bit more info about this approach https://tailwindcss.com/docs/extracting-components

app/assets/javascripts/ui_models/app_state/preferences_state.ts
```diff
@@ -0,0 +1,26 @@
+import { action, computed, makeObservable, observable } from 'mobx';
+
+export class PreferencesState {
+  private _open = true;
```
antsgar, line 4: We could make this one public and get rid of the getter

app/assets/javascripts/ui_models/app_state/preferences_state.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor() {
+    makeObservable<PreferencesState, '_open'>(this, {
+      _open: observable,
+      open: action,
+      close: action,
+      isOpen: computed,
+    });
+  }
+
+  open = (): void => {
```
antsgar, line 15: We may want to add a setter for this.open like this

setOpen(open: boolean): void {
  this.open = open;
}

and then we can either call that setter directly with true/false, or if you prefer to add two separate functions, maybe we could call them openPreferences or closePreferences to avoid confusion with the "open" attribute. Also not really important, but I think for state functions we're currently using named functions

app/assets/javascripts/components/IconButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+    className: '',
+  },
+  primary: {
+    className: 'info',
+  },
+};
+
+export type IconButtonType = keyof typeof ICON_BUTTON_TYPES;
+
+interface RoundButtonProps {
```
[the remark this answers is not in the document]
gorjan5sk: Missed this during renaming. Fixed.

app/assets/javascripts/preferences/index.tsx
```diff
@@ -0,0 +1,20 @@
+import { observer } from 'mobx-react-lite';
```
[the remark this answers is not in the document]
gorjan5sk: Done.

app/assets/stylesheets/_preferences.scss
gorjan5sk: Yep, fixed.

app/assets/javascripts/ui_models/app_state/preferences_state.ts
gorjan5sk: I wanted a more verby API to the property :) 
I adjusted the method names according to your suggestion below  :)

app/assets/javascripts/ui_models/app_state/preferences_state.ts
gorjan5sk: Done.

antsgar (approved):

PR #590, opened 2021-07-01
fix: fix tabindex of tag buttons

### LINKED ISSUE #589: Keyboard navigation broken with new tags
It's now hard to navigate from the title to the tag input and finally to the text editor.

tim-kilian:
Fixes #589 

I've added a negative tabIndex, so that the a tab will jump into the input field. From the input there is still the possibility to navigate the tab buttons with the arrow keys.

### DIFF
--- a/app/assets/javascripts/components/NoteTag.tsx
+++ b/app/assets/javascripts/components/NoteTag.tsx
@@ -87,6 +87,7 @@ export const NoteTag = observer(({ appState, tag }: Props) => {
       onKeyDown={onKeyDown}
       onFocus={onFocus}
       onBlur={onBlur}
+      tabIndex={-1}
     >
       <Icon type="hashtag" className="sn-icon--small color-info mr-1" />
       <span className="whitespace-nowrap overflow-hidden overflow-ellipsis max-w-290px">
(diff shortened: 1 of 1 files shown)


antsgar:
Hi @tim-kilian! Thank you for your contribution 🙂  I understand the problem, but would like to find a solution that preserves a more natural navigation order when using Tab, that is, avoiding the focus "jumping" to the input. What if we focused the first tag, and then on Tab press skipped the rest of them to focus the input? Then for tab navigation, we'd be considering all of the tags as whole container that could be "internally" navigated using arrow keys. What do you think?

mobitar:
I'm able to reproduce an issue where if I focus the title of a note with 1 tag, then try to tab my way to the note body, I'm unable to, as the tab gets stuck on the 1 tag.

antsgar:
> I'm able to reproduce an issue where if I focus the title of a note with 1 tag, then try to tab my way to the note body, I'm unable to, as the tab gets stuck on the 1 tag.

@mobitar correct, that was the bug fixed yesterday https://github.com/standardnotes/web/commit/2295f5d59364d8aaea4f9e818762807b2046beac (not yet released). However even with this fix the user will have to tab through all the tags before getting to the input.

antsgar:
I'm closing this one since I've opened a separate PR (#591) with an extension of this fix. Thank you so much for contributing and for reporting the issue!

PR #593, opened 2021-07-05
feat: enableUnfinishedFeatures flag and hidden preferences button

gorjan5sk:
- Implemented enableUnfinishedFeatures flag, applied as env var and cascades to AppState where it can be read
- Implemented rudimentary Preferences button to show/hide preferences screen, currently it fails silently since the preferences screen is in another PR

### DIFF
--- a/app/assets/javascripts/ui_models/app_state/app_state.ts
+++ b/app/assets/javascripts/ui_models/app_state/app_state.ts
@@ -48,8 +48,8 @@ export enum EventSource {
 type ObserverCallback = (event: AppStateEvent, data?: any) => Promise<void>;
 
 export class AppState {
-  readonly enableUnfinishedFeatures =
-    isDev || location.host.includes('app-dev.standardnotes.org');
+  readonly enableUnfinishedFeatures: boolean = (window as any)
+    ?._enable_unfinished_features;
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/ui_models/app_state/app_state.ts
```diff
@@ -80,6 +80,7 @@ export class AppState {
     application: WebApplication,
     private bridge: Bridge
   ) {
+    console.log('enableUnfinishedFeatures', this.enableUnfinishedFeatures);
```
mobitar, line 83:
```suggestion
```

gorjan5sk: Fixed.

antsgar: Should we maybe only print this log on the dev app?

antsgar (approved):

app/assets/javascripts/ui_models/app_state/app_state.ts
gorjan5sk: I removed that, we don't need it. :D I'm gonna resolve this conversation.

PR #600, opened 2021-07-14
feat: two factor authentication segment in preferences

gorjan5sk:
- Implement two factor authentication segment in Preferences screen, Security tab
- Currently it uses a mocked state implementation providing fake values.


app/assets/javascripts/preferences/view.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export interface PreferencesWrapperProps {
+  appState: { preferences: { isOpen: boolean; closePreferences: () => void } };
+}
+
+export const PreferencesViewWrapper: FunctionComponent<PreferencesWrapperProps> =
+  observer(({ appState }) => {
+    if (!appState.preferences.isOpen) return null;
+    return (
+      <PreferencesView close={() => appState.preferences.closePreferences()} />
```
antsgar, line 79: The idea behind having shared state is that different components can access it without passing down props to deeply nested components. In this case, we'd just pass appState, instead of passing close

app/assets/javascripts/preferences/view.tsx
```diff
@@ -0,0 +1,81 @@
...
+          />
+        </TitleBar>
+        <PreferencesCanvas preferences={prefs} />
+      </div>
+    );
+  }
+);
+
+export interface PreferencesWrapperProps {
+  appState: { preferences: { isOpen: boolean; closePreferences: () => void } };
```
antsgar, line 72:
```suggestion
  appState: AppState;
```

app/assets/javascripts/preferences/view.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <Title className="text-lg">Your preferences for Standard Notes</Title>
+          <RoundIconButton
+            onClick={() => {
+              close();
+            }}
+            type="normal"
+            icon="close"
+          />
+        </TitleBar>
+        <PreferencesCanvas preferences={prefs} />
```
antsgar, line 65: preferences is shared between PreferencesView, PreferencesCanvas and PaneSelector. Instead of passing it down to the deeply nested PaneSelector, it should be part of Mobx shared state.

app/assets/javascripts/preferences/panes/security.tsx
```diff
@@ -0,0 +1,13 @@
+import { observer } from 'mobx-react-lite';
+import { FunctionComponent } from 'preact';
+import { PreferencesPane } from '../components';
+import { Preferences } from '../models';
+import { TwoFactorAuthComponent } from './two-factor-auth';
+
+export const Security: FunctionComponent<{ prefs: Preferences }> = observer(
+  ({ prefs }) => (
+    <PreferencesPane>
+      <TwoFactorAuthComponent tfAuth={prefs.twoFactorAuth} />
```
antsgar, line 10: Same here, instead of reading twoFactorAuth from prefs, then passing it down to TwoFactorAuthComponent, TwoFactorAuthEnabled and TwoFactorAuthDisabled, all components who need access to it should read it from the shared state

app/assets/javascripts/preferences/components/pane.tsx
```diff
@@ -0,0 +1,35 @@
...
+        ))}
+  </div>
+);
+
+export const PreferencesPane: FunctionComponent = ({ children }) => (
+  <div className="preferences-pane flex-grow flex flex-row overflow-y-auto min-h-0">
+    <div className="flex-grow flex flex-col py-6 items-center">
+      <div className="w-125 max-w-125 flex flex-col gap-3">{children}</div>
+    </div>
+    <div className="flex-basis-55 flex-shrink-max" />
```
antsgar, line 33: This comment belongs in a StyleKit PR, but why do we need to set flex-shrink to 10000? Won't a flex-shrink with flex-shrink set to 1 do?

app/assets/javascripts/preferences/view.tsx
gorjan5sk: I'm intentionally decoupling the Preferences here from appState. It'll allow for easier testing of the views, the preferences state and in the future it won't block any potential refactors of AppState and it's dependencies.

app/assets/javascripts/preferences/components/pane.tsx
gorjan5sk: Yep, you're right. I was overcompensating for the feature I wanted :D Fixed now.

app/assets/javascripts/preferences/view.tsx
gorjan5sk: In general, I'm trying to reduce the surface API of React components so that they are more maleable, testable, isolated. Additionally, the current tight coupling between Preferences and TwoFactorAuthentication is only because the TwoFactorAuthentication is a "mocked" state - it provides hardcoded values and will be rewritten upon the integration with the whole system. The decoupling, as I mentioned before, will allow us not only to test the views (which isn't that important), it'll allow us to more precisely test the underlying state management (which is very important).

app/assets/javascripts/preferences/components/menu-item.tsx
```diff
@@ -0,0 +1,27 @@
+import { Icon, IconType } from '@/components/Icon';
```
antsgar, line 1: For components, let's name the file in camelcase (MenuItem in this case), to keep it consistent with the rest of the project

app/assets/javascripts/preferences/models/preferences.ts
```diff
@@ -0,0 +1,76 @@
+import { IconType } from '@/components/Icon';
```
antsgar, line 1: This model is pretty similar to the rest of the models on the shared state, so I think it'd make more sense to keep it in appState too

app/assets/javascripts/preferences/models/two-factor-auth.ts
```diff
@@ -0,0 +1,81 @@
+import { makeAutoObservable, observable } from 'mobx';
```
antsgar, line 1: Same here, why not keep this model in appState like the rest of them?

antsgar (approved):

PR #605, opened 2021-07-22
feat(preferences): 2FA activation dialog with mocked state

gorjan5sk:
- Implemented two factor authentication activation process in preferences
- Uses hard-coded values for interface, can be completed successfully without providing secret key and auth code


app/assets/javascripts/components/Button.tsx
```diff
@@ -0,0 +1,28 @@
+import { FunctionComponent } from 'preact';
+
+const base = `rounded px-4 py-1.75 font-bold text-sm fit-content cursor-pointer`;
+
+const normal = `${base} bg-default color-text border-solid border-gray-300 border-1 \
```
antsgar, line 5: Maybe we could call these baseClass, normalClass and primaryClass?

app/assets/javascripts/preferences/PreferencesView.tsx
```diff
@@ -37,19 +32,27 @@ const PaneSelector: FunctionComponent<{
...
+
+const PreferencesView: FunctionComponent<{ close: () => void }> = observer(
   ({ close }) => {
-    const prefs = new Preferences();
+    const prefs = new PreferencesMenu();
+
     return (
-      <div className="sn-full-screen flex flex-col bg-contrast z-index-preferences">
+      <div
+        onClick={blockPropagation}
```
antsgar, line 53: Could you explain a bit why we need this onClick since it's only stopping propagation?

app/assets/javascripts/preferences/panes/two-factor-auth/AuthAppInfoPopup.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * AuthAppInfoPopup is an info icon that shows a tooltip when clicked
+ * Tooltip is dismissible by clicking outside
+ *
+ * Note: it can be generalized but more use cases are required
+ * @returns
+ */
+export const AuthAppInfoTooltip: FunctionComponent = () => {
+  const [shown, setShow] = useState(false);
```
antsgar, line 13:
```suggestion
  const [shown, setShown] = useState(false);
```

Not required of course, I just find it easier to look for stuff if the update function is named after the state variable

app/assets/javascripts/preferences/panes/two-factor-auth/AuthAppInfoPopup.tsx
```diff
@@ -0,0 +1,44 @@
...
+    };
+  }, [ref]);
+
+  return (
+    <div className="relative">
+      <IconButton
+        icon="info"
+        className="mt-1"
+        onClick={() => {
+          setShow(!shown);
```
antsgar, line 30: For buttons that control visibility of another element, we use ReachUI's Disclosure https://reach.tech/disclosure. This ensures that we handle accessibility in the best possible way with a quick implementation

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -0,0 +1,135 @@
...
+      href="https://standardnotes.com/help/21/where-should-i-store-my-two-factor-authentication-secret-key"
+    >
+      Learn more
+    </a>
+  </Text>
+);
+
+export const TwoFactorActivationView: FunctionComponent<{
+  activation: TwoFactorActivation;
+}> = observer(({ activation: act }) => (
```
antsgar, line 127: Not sure if this is still changing on the next PR, but components not reading observables directly but receiving them as props don't need to be wrapped in observers

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -0,0 +1,135 @@
...
+    <PreferencesSegment>
+      <div className="flex flex-row items-center">
+        <div className="flex-grow flex flex-col">
+          <Title>Two-factor authentication</Title>
+          <Text>
+            An extra layer of security when logging in to your account.
+          </Text>
+        </div>
+        <Switch
+          checked={auth.enabled !== false}
```
antsgar, line 34: Can auth.enabled be anything different from a boolean?

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -0,0 +1,135 @@
+import { FunctionalComponent, FunctionComponent } from 'preact';
```
antsgar, line 1: I'm feeling the files to be a bit too cluttered by having many components on each one of them. Can we try to decompose them a bit more?

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -0,0 +1,135 @@
...
+      Learn more
+    </a>
+  </Text>
+);
+
+export const TwoFactorActivationView: FunctionComponent<{
+  activation: TwoFactorActivation;
+}> = observer(({ activation: act }) => (
+  <>
+    {act.step === 'scan-qr-code' && <ScanQRCode activation={act} />}
```
antsgar, line 129: Perhaps we could implement this with a switch?

app/assets/javascripts/preferences/panes/two-factor-auth/model.ts
```diff
@@ -0,0 +1,169 @@
+import { action, makeAutoObservable, observable, untracked } from 'mobx';
```
antsgar, line 1: These models are implemented using Mobx, any specific reason to keep them outside of the shared store?

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -0,0 +1,135 @@
...
+    />
+  );
+  const copy = (
+    <IconButton
+      icon="copy"
+      onClick={() => {
+        navigator?.clipboard?.writeText(secretKey);
+      }}
+    />
+  );
```
vardan-arm, line 93: We could create a separate higher order component that returns receives icon and click handler and returns corresponding `<IconButton />` component.

app/assets/javascripts/components/Button.tsx
gorjan5sk: Sure! Fixed in 311f50d00cef43129e5fa08819e9987a3625ee87.

app/assets/javascripts/preferences/PreferencesView.tsx
gorjan5sk: We don't, I missed this, it's a remnant of a bug I was fixing. Removed in 311f50d00cef43129e5fa08819e9987a3625ee87 .

app/assets/javascripts/preferences/panes/two-factor-auth/AuthAppInfoPopup.tsx
gorjan5sk: Changed. I also renamed `shown` to `isShown` for better clarity. 311f50d00cef43129e5fa08819e9987a3625ee87

app/assets/javascripts/preferences/panes/two-factor-auth/AuthAppInfoPopup.tsx
gorjan5sk: Done in 311f50d00cef43129e5fa08819e9987a3625ee87.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: It's true that the `activation` parameter is an observable for the outer React component, but it also contains observables itself which `TwoFactorActivationView` listens to, so it has to be an observer in order to react to those changes. If we remove the `observer` decoration, the dialog flow buttons won't react any more.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: Yes. And yes, it doesn't look good :) I'm in the process of changing that.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: But, isn't that what IconButton already does?

vardan-arm: I was referring to code duplication related to `IconButton`. For example, we have defined `copy` variable in 3 different places and only `onClick` handler is different there, so we could add abstraction over it. However, after looking into `IconButton` and its children, it seems introducing a new component will really not be so useful here.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: It's the same effect, but this way felt more natural since we're developing a template. It's more flexible to manipulation, e.g. we can add an extra condition, we can add decorations around it, if needed, stuff like that. I'd leave it as is if it's all the same to you.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: Done. Please verify since it's kind of subjective.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: It should be better now. Commit 1f75df8e6c5ab944f885face09ba067836916592.

app/assets/javascripts/preferences/panes/two-factor-auth/model.ts
gorjan5sk: Yes, the single responsibility principle. I don't want to clutter the shared state with all possible states the Preferences and 2FA can be in. 
Also, mobx observables don't need to be grouped in a single store. However, the reason it's not connected to it at all ATM is because the following PR is the one for integration with the actual activation.

antsgar: You’d just add an extra domain store to the root store, so it would still be isolated enough that it wouldn’t be cluttered. The advantage of having all stores grouped in a root one is they can share references and interact with each other if they need to.

gorjan5sk: I understand that advantage, but by exposing the whole state to any of the other components we are opening up to many bugs. If we share all the states it'd be pretty easy any component from anywhere to introduce an unexpected functionality. Additionally the outsider would have to handle all the possible states of the store that it uses, and it doesn't need to do that.
The simpler way, less bug-prone, would be to introduce a channel(method) towards the state that we want to access, and let the state handle the reception properly, in the exact place where we manage it! :D

antsgar: I’m not sure I agree tbh, I feel isolating each domain object from each other completely will make it difficult to share state between them in the future. So far I’ve found multiple domain stores grouped in a root store is a nice balance between having several stores each with their own responsibility but them being able to interact. I’d agree with this approach only if these models won’t need to have any interaction with the other stores, is this the case here?

gorjan5sk: > I’m not sure I agree tbh, I feel isolating each domain object from each other completely will make it difficult to share state between them in the future.

Not difficult at all. In fact, it's easier to reason about than handling all the combinations of a shared state in a variety of components. As I mentioned before: 
- If we want to trigger some event in the local state, we just implement a method which is called through an official channel instead of mutating the values directly. 
- If we want to make some properties readable by other components then we expose only those properties.

The alternative of a global shared state allows any component to change any state for any screen and that is a source of unpredictability. If multiple components alter and access the same property, there is no way of predicting what order the mutations will happen in, nor there is a way to guarantee that the changes will happen at the right moment.

gorjan5sk: It might even seem that it's more code, but it's not. Because we create an official channel to mutate the specific state, the outsider components need only call it, whereas if we mutate directly from those outsider components then they have to handle every intermediary state until the correct one is reached.
Not to mention that properly testing a several small states is much easier than testing one large shared state.

antsgar: @gorjan5sk I'll give this approach a try, I think it'll make more sense to me when I start applying it 🙂

antsgar (approved):

app/assets/javascripts/preferences/panes/two-factor-auth/model.ts
gorjan5sk: Cool! :) I suspect (but not sure) we'll see it in action in my next PR and we'll see how well it works in the current codebase.

PR #623, opened 2021-08-31
feat: pass web app version to snjs application

vardan-arm:


app/assets/javascripts/views/application/application_view.ts
```diff
@@ -81,6 +83,12 @@ class ApplicationViewCtrl extends PureViewCtrl<unknown, {
     await this.application.launch();
   }
 
+  async setAppVersion() {
+    const appVersion = packageJson.version;
```
mobitar, line 87: This value is already available globally: https://github.com/standardnotes/web/blob/develop/webpack.config.js#L17

vardan-arm: Didn't know that, thanks! Will update it.

vardan-arm:
Internal link - https://app.asana.com/0/1200588672119490/1200383926936362

app/assets/javascripts/ui_models/application.ts
```diff
@@ -25,6 +25,8 @@ import { NativeExtManager } from '@/services/nativeExtManager';
 import { StatusManager } from '@/services/statusManager';
 import { ThemeManager } from '@/services/themeManager';
 
+declare const __VERSION__: string;
```
mobitar, line 28: Can this be imported from somewhere instead? Perhaps we should have a version.ts file that all other files reference if they need to access the version?

vardan-arm: Nice point, moved this and other globals to separate files. Actually my intention was to combine them in one file end export, but only default export works in this case, so created files for every global item.

app/assets/javascripts/version.ts
```diff
@@ -0,0 +1,8 @@
+/** Declared in webpack config */
+declare const __VERSION__: string;
+declare const __DESKTOP__: boolean;
+declare const __WEB__: boolean;
+
+export const appVersion = __VERSION__;
```
mobitar, line 6: Hmm I would make these PascalCase to indicate a global constant rather than var or function.
```suggestion
export const AppVersion = __VERSION__;
```

vardan-arm: Correct, will update. Do you think we should make `isDesktopPlatform` and `isWebPlatform` Pascal case as well? I think based on their name it's better to keep them camel cased, but I'm open for other opinions :)

mobitar: Yup same thing for those as well since they are global constants.

vardan-arm: Updated)

karolsojko (approved):

antsgar (approved):

PR #624, opened 2021-09-01
chore: update snjs, sncrypto-web and fix tsc complaints

gorjan5sk:


app/assets/javascripts/views/footer/footer_view.ts
```diff
@@ -540,9 +540,11 @@ class FooterViewCtrl extends PureViewCtrl<
     const id = 'dock-svg-' + shortcut.component.uuid;
     const element = document.getElementById(id)!;
     const parser = new DOMParser();
-    const svg = shortcut.component.package_info.dock_icon.source;
-    const doc = parser.parseFromString(svg, 'image/svg+xml');
-    element.appendChild(doc.documentElement);
+    const svg = shortcut.component.package_info.dock_icon?.source;
+    if(svg!=null) {
```
mobitar, line 544: Be sure to set up auto-lint on save, or lint before committing.
```suggestion
    if (svg != null) {
```

app/assets/javascripts/views/footer/footer_view.ts
```diff
@@ -301,7 +301,7 @@ class FooterViewCtrl extends PureViewCtrl<
       CollectionSort.Title,
       'asc',
       (theme: SNTheme) => {
-        return theme.package_info && theme.package_info.dock_icon;
+        return theme.package_info && theme.package_info.dock_icon != null;
```
mobitar, line 304:
```suggestion
        return theme.package_info && !isNullOrUndefined(theme.package_info.dock_icon);
```

gorjan5sk: But, why a function for a language feature?

mobitar: What if it's equal to undefined here?

gorjan5sk: The `x != null`  checks for both `null` and `undefined`.
If I had used `x !== null` then it would only check for `null`.

mobitar: Ah ok. So why check for null instead of undefined if both are the same?

gorjan5sk: Just a habit.

mobitar: If both the same let's go with undefined, since that's far more prevalent in our codebases than null.

app/assets/javascripts/views/footer/footer_view.ts
antsgar: Should we check for undefined here too?

mobitar (approved):

PR #626, opened 2021-09-03
feat: integrate two factor authentication

gorjan5sk:
Integrate fetching, activation and disabling two factor authentication


app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,120 @@
...
+  (s as any).type === 'two-factor-activation';
+
+export const is2FAEnabled = (s: TwoFactorStatus): s is 'two-factor-enabled' =>
+  s === 'two-factor-enabled';
+
+export class TwoFactorAuth {
+  private _status: TwoFactorStatus | 'fetching' = 'fetching';
+  private _errorMessage: string | null;
+
+  constructor(private readonly mfa: MfaGateway) {
```
mobitar, line 23: The mfa property when read and used throughout this code really doesn't add any context as to what exactly the object is. MFA stands for "multi-factor authenticator" so to name an object that doesn't make too much sense.

```suggestion
  constructor(private readonly gateway: MfaGateway) {
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -28,26 +27,25 @@ export const TwoFactorAuthView: FunctionComponent<{
           <Text>
             An extra layer of security when logging in to your account.
           </Text>
+          {auth.errorMessage != null && (
+            <Text className="color-dark-red">{auth.errorMessage}</Text>
```
mobitar, line 31: We should be using StyleKit colors/classes. dark-red is subjective and relative to the theme. Should be SK `danger` class.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorActivation.ts
```diff
@@ -0,0 +1,125 @@
...
+  get activationStep() {
+    return this._activationStep;
+  }
+
+  get verificationStatus() {
+    return this._2FAVerification;
+  }
+
+  get qrCode() {
+    return `otpauth://totp/2FA?secret=${this._secretKey}&issuer=Standard%20Notes`;
```
mobitar, line 67: (we'd need to get user email into this function)

```suggestion
    return `otpauth://totp/2FA?secret=${this._secretKey}&issuer=Standard%20Notes&label=${user.email}`;
```

This wasn't present in MFA Link because it didn't have access to user email, but we do here.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: That's a different nuance of red.

mobitar: Nothing should be colored without StyleKit in general. Otherwise users can't customize appearance.

gorjan5sk: The 'color-dark-red' is defined in StyleKit. I understand what you mean, that the name should match the context. But seeing as danger was already taken, I couldn't find a different appropriate name that wouldn't overshadow danger's meaning. Shall I use "color-danger" instead, and we update the color schemes later?

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
gorjan5sk: Called it `mfaGateway`. Resolving comment.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: Used color-danger :)

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorActivation.ts
gorjan5sk: Done.

app/assets/javascripts/preferences/panes/two-factor-auth/ScanQRCode.tsx
```diff
@@ -35,7 +38,7 @@ export const ScanQRCode: FunctionComponent<{
       <TwoFactorDialogDescription>
         <div className="flex flex-row gap-3 items-center">
```
antsgar, line 39: Have you tested the layout on the desktop application? I have recently registered a styling bug due to the `gap-x` classes, seems they're not working properly on Electron

app/assets/javascripts/preferences/PreferencesWrapper.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { PreferencesView } from './PreferencesView';
+
+export interface PreferencesWrapperProps {
+  appState: { preferences: { isOpen: boolean; closePreferences: () => void } };
+  application: WebApplication;
+}
+
+export const PreferencesViewWrapper: FunctionComponent<PreferencesWrapperProps> =
+  observer(({ appState, application }) => {
+    if (!appState.preferences.isOpen) return null;
```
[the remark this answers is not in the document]
gorjan5sk: Done.

app/assets/javascripts/preferences/components/Pane.tsx
```diff
@@ -16,12 +16,14 @@ export const PreferencesGroup: FunctionComponent = ({ children }) => (
...
-          </>
-        ))}
+      : children
+          .filter((c) => !!c)
+          .map((c, i, arr) => (
+            <>
+              {c}
+              <HorizontalLine index={i} length={arr.length} />
+            </>
+          ))}
```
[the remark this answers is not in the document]
gorjan5sk: Done :)

app/assets/javascripts/preferences/components/Pane.tsx
```diff
@@ -16,12 +16,14 @@ export const PreferencesGroup: FunctionComponent = ({ children }) => (
...
     {!Array.isArray(children)
       ? children
-      : children.map((c, i, arr) => (
-          <>
-            {c}
-            <HorizontalLine index={i} length={arr.length} />
-          </>
-        ))}
+      : children
+          .filter((c) => !!c)
```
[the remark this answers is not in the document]
gorjan5sk: I've changed the filter to `c!=null && c!==''` to be more descriptive. Basically since we're putting a HorizontalLine between segments, and the children array contains null children it'll put a horizontal line between them because they're members of the array.

app/assets/javascripts/preferences/panes/Security.tsx
```diff
@@ -1,9 +1,10 @@
 import { FunctionComponent } from 'preact';
 import { PreferencesPane } from '../components';
 import { TwoFactorAuthWrapper } from './two-factor-auth';
+import { MfaProps } from './two-factor-auth/MfaProps';
 
-export const Security: FunctionComponent = () => (
+export const Security: FunctionComponent<MfaProps> = (props) => (
   <PreferencesPane>
-    <TwoFactorAuthWrapper />
+    <TwoFactorAuthWrapper {...props} />
```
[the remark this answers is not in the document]
gorjan5sk: Since props spreading is destructuring the props the children will only use the props they subscribe to. This code is equivalent to:
```
// Declaring and Initializing to better illustrate
interface PreferencesProps {
  closePreferences:()=>void;
  application:WebApplication;
  mfaGateway:MfaGateway;
}
const prefProps = {...randomPrefProps};

// this is the line equivalent to passing the props to the component.
const { mfaGateway } = prefProps;
```

So basically the child only acquires the props it needs.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorActivation.ts
```diff
@@ -0,0 +1,126 @@
...
+        _enable2FAVerification: observable,
+        refreshOtp: action,
+        inputOtpToken: observable,
+        inputSecretKey: observable,
+      },
+      { autoBind: true }
+    );
+  }
+
+  get secretKey() {
```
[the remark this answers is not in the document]
gorjan5sk: Done.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,120 @@
...
+  s === 'two-factor-disabled';
+
+export const is2FAActivation = (s: TwoFactorStatus): s is TwoFactorActivation =>
+  (s as any).type === 'two-factor-activation';
+
+export const is2FAEnabled = (s: TwoFactorStatus): s is 'two-factor-enabled' =>
+  s === 'two-factor-enabled';
+
+export class TwoFactorAuth {
+  private _status: TwoFactorStatus | 'fetching' = 'fetching';
```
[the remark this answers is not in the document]
gorjan5sk: Because `fetching` is not a status that the two factor authentication can be in. If you see the getter below, you'll see that `fetching` is reduced to a disabled status. If we add a spinner for an in-flight request we'd probably expose another getter as well.
Functionally it's the same, but this has a more descriptive purpose - to say that "fetching is not a TwoFactorStatus".

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,120 @@
...
+      '_status' | '_errorMessage' | 'deactivateMfa' | 'startActivation'
+    >(this, {
+      _status: observable,
+      _errorMessage: observable,
+      deactivateMfa: action,
+      startActivation: action,
+    });
+  }
+
+  private startActivation() {
```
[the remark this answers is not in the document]
gorjan5sk: Done.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  setError(errorMessage: string | null) {
+    this._errorMessage = errorMessage;
+  }
+
+  toggle2FA() {
+    if (!this.isLoggedIn) return;
+    if (this._status === 'two-factor-disabled') return this.startActivation();
+    if (this._status === 'two-factor-enabled') return this.deactivate2FA();
+  }
```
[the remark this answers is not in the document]
gorjan5sk: Done.

app/assets/javascripts/preferences/panes/two-factor-auth/Verification.tsx
```diff
@@ -57,7 +63,7 @@ export const Verification: FunctionComponent<{
           className="min-w-20"
           type="primary"
           label="Next"
-          onClick={() => act.enable2FA('X', 'X')}
+          onClick={() => act.enable2FA()}
```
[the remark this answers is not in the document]
gorjan5sk: Done, on all of the onClick handlers here.

app/assets/javascripts/preferences/panes/two-factor-auth/ScanQRCode.tsx
gorjan5sk: For some reason the build on desktop is failing. The errors I'm seeing are webpack css failures connected with reach components. I think that needs to be resolved as part of another task.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorActivation.ts
```diff
@@ -0,0 +1,127 @@
+import { action, makeAutoObservable, observable, untracked } from 'mobx';
+import { MfaGateway } from './MfaProps';
+
+const activationSteps = [
+  'scan-qr-code',
+  'save-secret-key',
+  'verification',
+] as const;
```
vardan-arm, line 8: Maybe we could store activation steps in an enum and use its values instead of the strings. In that case we could avoid defining new `ActivationStep` type at line 10 and just use that enum as _activationStep's type.

app/assets/javascripts/preferences/components/PreferencesPane.tsx
```diff
@@ -14,14 +14,16 @@ export const PreferencesSegment: FunctionComponent = ({ children }) => (
...
-    {!Array.isArray(children)
+    {Array.isArray(children)
       ? children
-      : children.map((c, i, arr) => (
-          <>
-            {c}
-            <HorizontalLine index={i} length={arr.length} />
-          </>
-        ))}
+          .filter((c) => c != null && c !== '')
```
vardan-arm, line 19: Nitpick, but let's explicitly use strict equality comparisons. And maybe more descriptive variable names.
```suggestion
          .filter((child) => child !== null && child !== '')
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,128 @@
...
+      )
+      .catch(
+        action((e) => {
+          this.setError(e.message);
+        })
+      );
+  }
+
+  private get isLoggedIn(): boolean {
+    return this.mfaGateway.getUser() != undefined;
```
vardan-arm, line 75:
```suggestion
    return this.mfaGateway.getUser() !== undefined;
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -28,26 +27,25 @@ export const TwoFactorAuthView: FunctionComponent<{
           <Text>
             An extra layer of security when logging in to your account.
           </Text>
+          {auth.errorMessage != null && (
```
vardan-arm, line 30:
```suggestion
          {auth.errorMessage !== null && (
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -0,0 +1,128 @@
...
+      return this.deactivate2FA();
+    }
+  }
+
+  get errorMessage(): string | null {
+    return this._errorMessage;
+  }
+
+  get status(): TwoFactorStatus {
+    if (this._status === 'fetching') return 'two-factor-disabled';
```
vardan-arm, line 125: Let's fix this one-liner `if` too:
```suggestion
    if (this._status === 'fetching') {
      return 'two-factor-disabled';
    }
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorActivation.ts
gorjan5sk: Ah, actually we don't need the list of steps any more. Changed to string literal type.

app/assets/javascripts/preferences/components/PreferencesPane.tsx
gorjan5sk: Changed c to child, but the comparison `!=null` compares both whether the variable is null or undefined, so it's fine like that. That's why the TS compiler doesn't complain.
```
null != null => false;
undefined != null => false;
```

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
gorjan5sk: Same as above.

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
gorjan5sk: ...

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
gorjan5sk: Done.

app/assets/javascripts/preferences/panes/Security.tsx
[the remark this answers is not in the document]
antsgar: . If the `Security` component has X props and you use props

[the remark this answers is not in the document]
antsgar: > So basically the child only acquires the props it needs.

This isn't really true. If the Security component has X props and you use props spreading to pass them down to TwoFactorAuthWrapper, TwoFactorAuthWrapper will get all X props whether it uses them or not. Per React Docs:

> Spread attributes can be useful but they also make it easy to pass unnecessary props to components that don’t care about them or to pass invalid HTML attributes to the DOM. We recommend using this syntax sparingly.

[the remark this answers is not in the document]
gorjan5sk: Ah, yeah, that's true. I conflated the spread with the destructuring.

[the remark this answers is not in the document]
gorjan5sk: Fixed!

app/assets/javascripts/preferences/components/PreferencesPane.tsx
vardan-arm: Hmm, in that case I would prefer to compare against both `null` and `undefined` explicitly or have some utility function which does that in case there are many such checks. 
Having written it explicitly would actually improve the readability, also the code will be more consistent, since here we are checking the value against `null`, but in `TwoFactorAuth` - against `undefined`.

gorjan5sk: There is a utility function in snjs already, but it's a feature of the language, it's part of the syntax, we don't need a utility function. I've changed it to undefined, for consistency.

app/assets/javascripts/preferences/panes/Security.tsx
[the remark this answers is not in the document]
antsgar: Awesome! 🙂 can we do the same for props spreading on `PreferencesCanvas` and `PaneSelector`?

antsgar (approved):

app/assets/javascripts/preferences/panes/Security.tsx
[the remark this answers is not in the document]
gorjan5sk: Those components are gonna need everything but `closePreferences` so I'd leave them as they are. They're more of pass-through layout components which are gonna be responsible to handing over the appropriate props to the chidren.

vardan-arm (approved):

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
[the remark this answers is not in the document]
mobitar: Linter or TS comiler should be throwing warnings for lack of return types, so if that's not happening in VSCode for you make sure it's set up properly. I think based on previous feedback your linter is not properly configured.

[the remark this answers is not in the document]
gorjan5sk: Thanks! I'll look into it.

PR #631, opened 2021-09-07
feat: disabled 2fa feature

gorjan5sk:
- use new snjs mfa method to check if 2fa feature is available


app/assets/javascripts/preferences/PreferencesViewWrapper.tsx
```diff
@@ -18,7 +18,8 @@ export const PreferencesViewWrapper: FunctionComponent<PreferencesViewWrapperPro
       <PreferencesView
         closePreferences={() => appState.preferences.closePreferences()}
         application={application}
-        mfaGateway={application}
+        mfaProvider={application}
+        userProvider={application}
```
vardan-arm, line 22: We use 2 separate types for `mfaProvider` and `userProvider` but pass the same `application` object to them. Probably it's better to either use one prop for `application` with appropriate name or pass application's respective fields to `mfaProvider` and `userProvider` (e.g., pass only `application.getUser` to `userProvider`)

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
```diff
@@ -11,7 +11,7 @@ export const is2FADisabled = (s: TwoFactorStatus): s is 'two-factor-disabled' =>
   s === 'two-factor-disabled';
 
 export const is2FAActivation = (s: TwoFactorStatus): s is TwoFactorActivation =>
```
vardan-arm, line 13: Let's use more meaningful name for `s` :)

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuthView.tsx
```diff
@@ -7,45 +7,38 @@ import {
...
+        <div className="flex flex-row items-center">
+          <div className="flex-grow flex flex-col">
+            <Title>Two-factor authentication</Title>
+            <Text>
+              An extra layer of security when logging in to your account.
+            </Text>
+          </div>
+          <Switch
+            checked={!is2FADisabled(auth.status)}
+            onChange={() => auth.toggle2FA()}
```
vardan-arm, line 28:
```suggestion
            onChange={auth.toggle2FA}
```

app/assets/javascripts/preferences/PreferencesViewWrapper.tsx
gorjan5sk: Separation of concerns. We can test  components that only depend on x:getUser(), that only depend on MfaProvider stuff, without mocking the whole WebApplication along with the kitchen sink :D

app/assets/javascripts/preferences/panes/two-factor-auth/TwoFactorAuth.ts
gorjan5sk: Done.

app/assets/javascripts/preferences/PreferencesViewWrapper.tsx
vardan-arm: It looks very strange tbh :D But I think we'll have better (practical) understanding of pros and cons of this approach only after writing some tests :)

gorjan5sk: They're coming! :D 
Can I get an approval now? :)

vardan-arm (approved):

PR #634, opened 2021-09-10
feat: add changing email

karolsojko:
Internal link: https://app.asana.com/0/1199914526147392/1200887853314815/f


vardan-arm:
Probably we could combine the logic that is the same in "change email" and "change password" in one place, but feel free to merge it this way now.

vardan-arm (approved):

PR #636, opened 2021-09-14
feat: implement delete account pane logic

vardan-arm:


vardan-arm:
Internal link - https://app.asana.com/0/1200588672119490/1200939057521441

app/assets/javascripts/enums.ts
```diff
@@ -22,3 +22,16 @@ export enum HtmlInputTypes {
   Url = 'url',
   Week = 'week'
 }
+
+export enum ErrorMessages {
```
mobitar, line 26: Nitpick but enums should typically be singular :) Same for HtmlInputTypes above but that's out of scope.

```suggestion
export enum ErrorMessage {
```

mobitar: In this case though it kind of reads weird when importing so maybe we can leave it as is. 

import { ErrorMessage } from '@/enums';

app/assets/javascripts/preferences/panes/account/deleteAccountDialog/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    if (!success) {
+      snAlert(IncorrectPassword);
+      cleanupBeforeCancelSubmit();
+      return;
+    }
+
+    const userEmail = (application.getUser() as User).email;
+
+    try {
+      const res = await fetch(`https://api.standardnotes.com/v1/reset`, {
```
mobitar, line 82: Hmm, this should be part of SNJS. Clients shouldn't make API requests directly. Especially because these requests will need to be authenticated.

app/assets/javascripts/preferences/panes/account/deleteAccountDialog/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    if (currentStep === Steps.FinishStep) {
+      onCloseDialog();
+      return;
+    }
+
+    setIsRequestInProgress(true);
+
+    if (password === '') {
```
mobitar, line 66:
```suggestion
    if (password.trim() === '') {
```

vardan-arm: I implemented it this way, but then removed, since I thought specifically for passwords we don't need trimming, as the user might put space in their password. I know it's not happening every day, but it's something we might consider, wdyt?

mobitar: Ah true 👍

app/assets/javascripts/enums.ts
vardan-arm: Yeah, it really could be a bit confusing when is singular. But I guess it's better to have consistency and use best practices here, so I renamed it and HtmlInputTypes too, which I added in the last PR.

vardan-arm:
Dependent on this PR - https://github.com/standardnotes/snjs/pull/432/files

gorjan5sk (approved):

gorjan5sk:
There are conflicts that need to be resolved, but otherwise, looks good.

PR #643, opened 2021-09-22
fix: encryption status title

gorjan5sk:


mobitar (changes requested):
Needs subtitle when no encryption available.

gorjan5sk:
Ah, there is a status, but there's also a bug. I'm reusing the status from the account menu, but it's not set unless the account menu has been opened. I'm on it @mobitar .

gorjan5sk:
Ok, now it should be reliable.

app/assets/javascripts/preferences/panes/Encryption.tsx
```diff
@@ -35,14 +36,25 @@ const EncryptionEnabled: FunctionComponent<{ appState: AppState }> = observer(({
   );
 });
 
-export const EndToEndEncryption: FunctionComponent<{ appState: AppState }> = observer(({ appState }) => {
+export const Encryption: FunctionComponent<{ appState: AppState }> = observer(({ appState }) => {
+  const app = appState.application;
+  const hasUser = app.hasAccount();
+  const hasPasscode = app.hasPasscode();
+  const isEncryptionEnabled = hasUser || hasPasscode;
```
mobitar, line 43:
```suggestion
  const isEncryptionEnabled = app.isEncryptionAvailable();
```

mobitar (approved):

PR #651, opened 2021-09-28
feat: Add Listed pane in preferences

amanharwara:
This PR adds a new preferences pane for integration with Listed. When there are no current Listed installations, the pane shows information about Listed and how to get started:

Listed-Pane-Empty

When there are present Listed installations, it shows a list of the blogs with options to open the blog, open the settings and disconnect the blog:

Listed-Pane-WithItems

When the disconnect button is pressed, it shows a confirmation dialog before disconnecting:

Listed-Pane-Disconnect


mobitar:
@amanharwara please pull in standardnotes:develop to your branch to resolve conflicts :)

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
```diff
@@ -0,0 +1,79 @@
+import { Button } from '@/components/Button';
+import { HorizontalSeparator } from '@/components/shared/HorizontalSeparator';
+import { LinkButton, Subtitle } from '@/preferences/components';
+import { WebApplication } from '@/ui_models/application';
+import { Action, SNComponent, SNItem } from '@standardnotes/snjs/dist/@types';
+import { JSXInternal } from 'preact/src/jsx';
+import React, { useState } from 'react';
```
gorjan5sk, line 7: Since we're using Preact, it's fine to delete this line and just import `useState` from preact.
Additionally, `<React.Fragment> ... </React.Fragment>` can be replaced by `<> ... </>`

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+  application: WebApplication;
+};
+
+export const BlogItem = ({
+  item,
+  showSeparator,
+  disabled,
+  disconnect,
+  application,
+}: Props): JSXInternal.Element => {
```
gorjan5sk, line 23: Instead of specifying the output type, it's better to specify that BlogItem is a FunctionComponent, like so:
`export const BlogItem: FunctionComponent<Props> = ...`
This way, the `item` property's type will be inferred and all the `(item as any)` casts can be removed.

app/assets/javascripts/preferences/panes/Listed.tsx
```diff
@@ -0,0 +1,110 @@
...
+  Title,
+  Subtitle,
+  Text,
+  LinkButton,
+} from '../components';
+import { observer } from 'mobx-react-lite';
+import { WebApplication } from '@/ui_models/application';
+import { ContentType, SNComponent } from '@standardnotes/snjs';
+import { SNItem } from '@standardnotes/snjs/dist/@types/models/core/item';
+import { useCallback, useEffect, useState } from 'react';
```
gorjan5sk, line 14: Should be imported from preact.

gorjan5sk (commented):
Some small finesse changes needed for code consistency and type-safety, otherwise good work! :)

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
amanharwara: Removing the type casts gives me this error: `Property 'actions' does not exist on type 'ComponentPackageInfo'.`

amanharwara:
Made the change from react to preact hooks and `FunctionalComponent` instead of `JSXInternal.Element`.

app/assets/javascripts/preferences/panes/Listed.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const disconnectListedBlog = (item: SNItem) => {
+    setIsDeleting(true);
+    application
+      .deleteItem(item)
+      .then(() => {
+        reloadItems();
+        setIsDeleting(false);
+      })
+      .catch((err) => {
+        console.error(err);
```
antsgar, line 47: Would be good to add some UI here to indicate to the user that something has gone wrong

app/assets/javascripts/preferences/panes/Listed.tsx
```diff
@@ -0,0 +1,110 @@
...
+      });
+  };
+
+  return (
+    <PreferencesPane>
+      {items.length > 0 && (
```
antsgar, line 53: Perhaps we could use a ternary operator here like so `items.length > 0 ? ... : ...` so it's easier to read

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+          link={
+            (item as any).package_info.actions.find(
+              (action: Action) => action.label === 'Settings'
+            ).url
+          }
+        />
+        <Button
+          type="danger"
+          label={isDisconnecting ? 'Disconnecting...' : 'Disconnect'}
+          disabled={disabled}
```
antsgar, line 72: Maybe we could use the `isDiconnecting` state variable here and get rid of the `isDeleting` one on the parent component

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+        'Disconnecting will result in loss of access to your blog. Ensure your Listed author key is backed up before uninstalling.',
+        `Disconnect blog "${item.name}"?`,
+        'Disconnect',
+        1
+      )
+      .then((shouldDisconnect) => {
+        if (shouldDisconnect) {
+          disconnect(item);
+        } else {
+          setIsDisconnecting(false);
```
antsgar, line 41: `setIsDisconnecting(false)` would need to go in a `finally` block to ensure it executes even if there's an error

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <>
+      <Subtitle>{item.name}</Subtitle>
+      <div className="flex">
+        <LinkButton
+          className="mr-2"
+          label="Open Blog"
+          link={
+            (item as any).package_info.actions.find(
```
antsgar, line 55: Is it possible here to use a specific type here instead of `any`?

antsgar:
@amanharwara really nice job! 🙂

app/assets/javascripts/preferences/panes/Listed.tsx
amanharwara: An alert with the error would work?

app/assets/javascripts/preferences/panes/Listed.tsx
amanharwara: Sure, done.

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
amanharwara: Since the deletion happens asynchronously, I've kept the isDeleting on the parent so the other Disconnect buttons can stay disabled while one is disconnecting to avoid any errors.

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
amanharwara: Putting it in the `finally` block turns it off even before it finishes disconnecting, since the `deleteItem()` function is async. I've just put `setIsDisconnecting(false)` in the `catch` block to execute it even if there is an error.

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
mobitar: `item as SNComponent` most likely (or maybe SNExtension)

amanharwara: I've tried both `SNComponent` and `SNActionsExtension` but both give errors: `Property 'actions' does not exist on type 'ComponentPackageInfo'.`

The reason I'm using `package_info.actions` instead of accessing `actions` is because when I'm getting the items using the `getItems()` function, `actions` is empty and gives an error, whereas the actions can be found in `package_info.actions`. This is when the site is running.

Am I doing something wrong here?

amanharwara: `item.actions` is empty whereas `item.package_info.actions` has all the correct actions.

amanharwara: I tried using `application.actionsManager.getExtensions()` instead of `application.getItems()` as well, but still the `actions` array seems to be empty while `package_info.actions` has the correct actions.

mobitar: Hmm actually I just realized that actions are downloaded for a blog only when you open the Actions menu in the editor pane. So if you install a new Listed blog but never open the Actions menu, then item.package_info.actions will probably be empty as well.

In that case we'll want to do what the Actions menu is doing, which is downloading the actions for an extension, then displaying the Show Blog option.

amanharwara:
Hi everyone,
I reckon with the latest commit, this PR should be ready for merging.
Can someone re-review?

app/assets/javascripts/preferences/panes/listed/BlogItem.tsx
antsgar: Shouldn't we await `disconnect` in that case? Might be a good idea to refactor these to use `async/await` which is a bit easier to read I think

amanharwara: That would require the disconnect function to be promisified. That is actually how I had originally implemented it, but I thought I was overcomplicating it, so I removed the promise.

amanharwara: @antsgar So I've turned the disconnect function into a Promise and using `await disconnect()` in the BlogItem and `setIsDisconnecting()` inside the `.finally()` block.

antsgar (approved):

PR #654, opened 2021-09-28
feat: purchase page iframe

antsgar:


mobitar (approved):

gorjan5sk:
Code-wise it's great, guessing just needs an snjs update?

antsgar:
@gorjan5sk yup the SNJS update is what's missing! But I won't be merging this yet since we'll be making a change in direction and won't be using iframes for this. Will set it to draft for now

PR #655, opened 2021-09-29
feat: slim account menu and icon buttons

gorjan5sk:
- Slim account menu
- Icons for account menu and preferences


app/assets/javascripts/views/footer/footer-view.pug
```diff
@@ -1,28 +1,37 @@
...
+                class-name="hover:color-info w-5 h-5 max-h-5"
+              )
         account-menu(
           ng-click='$event.stopPropagation()',
           app-state='ctrl.appState'
           application='ctrl.application'
           ng-if='ctrl.showAccountMenu',
         )
-      .sk-app-bar-item(
+      .sk-app-bar-item.ml-0-important(
```
vardan-arm, line 25: I know it's sometimes not that easy, but could we try to avoid `important` class here? Maybe in this case we can make `.ml-0` more specific in corresponding `scss` file. No need to spend too much time on this anyway, probably just give it a shot.

vardan-arm:
Good work @gorjan5sk ! Just left one comment, also please remove unused variables (there are some warnings in this PR).

vardan-arm (approved):

app/assets/javascripts/views/footer/footer-view.pug
gorjan5sk: We can't avoid it here, the global selector is overwriting the property.

mobitar (approved):

PR #696, opened 2021-10-21
feat: remove extensions manager

atmoio:


VPaulV:
Hi guys, after extensions manager have been removed I cannot add my self hosted extensions with extensions.json file. Is there any alternative or now I should add every plugin one by one?

PR #733, opened 2021-11-11
feat: Pass period & plan params to purchase page

amanharwara:


app/assets/javascripts/purchaseFlow/PurchaseFlowWrapper.tsx
```diff
@@ -27,8 +29,11 @@ export const loadPurchaseFlowUrl = async (
   application: WebApplication
 ): Promise<boolean> => {
   const url = await getPurchaseFlowUrl(application);
+  const params = new URLSearchParams(window.location.search);
+  const period = params.get('period') ?? '';
+  const plan = params.get('plan') ?? '';
   if (url) {
-    window.location.assign(url);
+    window.location.assign(`${url}&period=${period}&plan=${plan}`);
```
mobitar, line 36: Feel uneasy about a url that could look like `.com/purchase?token=123&period=&plan=`. We should instead conditionally build the string based on availability of params.

vardan-arm (approved):

PR #739, opened 2021-11-15
fix: Remove HTML tags before counting words for a note

amanharwara:
Asana task: https://app.asana.com/0/1199991945165772/1201372131831438/f


app/assets/javascripts/components/NotesOptions.tsx
```diff
@@ -44,6 +44,10 @@ const countNoteAttributes = (text: string) => {
       paragraphs: 'N/A',
     };
   } catch {
+    const removeTags = document.createElement("div");
+    removeTags.innerHTML = text;
```
vardan-arm, line 48: The only concern I have is whether we aren't introducing a possibility of XSS attack here? Although I may have overthinking as we don't append the newly created `div` to the DOM.

amanharwara: So I tried this by adding 
```
<script>
alert(1);
</script>
```
to the contents of a Bold Editor note, and then right-clicked to get the note count. Didn't get any alert popping up.
I think the editors sanitize the text anyways, so it shouldn't be a problem.

vardan-arm: I also tried a few XSS scripts and they didn't show any popup as well.

> the editors sanitize the text anyways, so it shouldn't be a problem.

In this case we put the HTML content directly, bypassing editors' sanitization, but since we don't append the created div in any case, I think it should be safe to go.

vardan-arm (approved):

PR #750, opened 2021-11-24
fix: check whether `valid_until` is correct date before passing it further

vardan-arm:
Internal link: https://app.asana.com/0/1200588672119490/1201419221095307


vardan-arm:
Not required.

PR #758, opened 2021-12-01
feat: Nativize "No distraction" theme as "Focus Mode"

amanharwara:
Asana task: https://app.asana.com/0/0/1201366388874772/f


app/assets/stylesheets/_focused.scss
```diff
@@ -0,0 +1,62 @@
...
+  .section.notes {
+    flex: none !important;
+    width: 0px !important;
+    transition: width 0.25s;
+    border: none !important;
+  }
+
+  .section.notes:hover {
+    flex: initial;
+    width: 0px !important;
```
vardan-arm, line 59: Could we avoid using `important` in this file?

vardan-arm:
Looks good @amanharwara ! Only left a comment related to styles.

app/assets/stylesheets/_focused.scss
amanharwara: Unfortunately those will have to be kept as `.section.notes` and `.section.tags` have a `style` attribute in the DOM that will override these values and break the functionality of this.

vardan-arm (approved):

moughxyz (approved):

PR #759, opened 2021-12-03
fix: Width transition on manually resizing sections

amanharwara:
Transition should only happen when toggling Focus Mode, not when manually resizing sections.


moughxyz (approved):

app/assets/javascripts/components/QuickSettingsMenu/QuickSettingsMenu.tsx
```diff
@@ -42,7 +42,7 @@ const toggleFocusMode = (enabled: boolean) => {
       document.body.classList.remove('focus-mode');
       setTimeout(() => {
         document.body.classList.remove('disable-focus-mode');
-      }, 315);
+      }, 1255);
```
moughxyz, line 45: Can we move this into a named const that describes where the number came from? `const cssAnimationDuration`

moughxyz (changes requested):

PR #761, opened 2021-12-05
refactor: Move notes_view to React

amanharwara:


app/assets/javascripts/components/NotesList.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    const tagsStringForNote = (note: SNNote): string => {
+      if (hideTags) {
+        return '';
+      }
+      const selectedTag = appState.selectedTag;
+      if (!selectedTag) {
+        return '';
+      }
+      if (selectedTag?.isSmartTag) {
```
laurentsenta, line 34: Can you remove the `?` check here?

Typescript should be smart enough to know there is a value (since we would have exited the code with the return before),
if it doesn't then we might have missed a case.

app/assets/javascripts/components/NotesList.tsx
```diff
@@ -0,0 +1,118 @@
...
+          .getNoteTags(note)
+          .map((tag) => '#' + tag.title)
+          .join(' ');
+      }
+      const tags = appState.getNoteTags(note);
+      if (tags.length === 1) {
+        return '';
+      }
+      return tags.map((tag) => '#' + tag.title).join(' ');
+    };
```
laurentsenta, line 45: Something looks fishy with this code, it's probably on me but it looks like we're saying:

- if we have a smart tag: we dump the list of tags
- if we have a regular tag: we ignore the list of tags is there is only one, else we dump the list of tags.

I'd assume both case should do the same, maybe:

- if we have a smart tag: we ignore the list of tags is there is only one, else we dump the list of tags.
- if we have a regular tag: we ignore the list of tags is there is only one, else we dump the list of tags.

Could we do something like:

```ts
const tagTitleWithHash = tag => `#${tag.title}`
appState
   .getNoteTags(note)
   .filter(tag => tag.uuid !== selectedTag.uuid)
   .map(tagTitleWithHash)
   .join(' ')
```

?

app/assets/javascripts/components/NotesList.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+    const onContextMenu = (note: SNNote, posX: number, posY: number) => {
+      if (!selectedNotes[note.uuid]) {
+        appState.notes.selectNote(note.uuid, true);
+        openNoteContextMenu(posX, posY);
+      } else {
+        openNoteContextMenu(posX, posY);
+      }
+    };
```
laurentsenta, line 63: Could you use something like:

```ts
      if (!selectedNotes[note.uuid]) {
        appState.notes.selectNote(note.uuid, true);
      }
      openNoteContextMenu(posX, posY);
```

to make the intent more clear? (select if not already, then open context menu)

or even:

```ts
      appState.notes.selectNote(note.uuid, true);
      openNoteContextMenu(posX, posY);
```

(always select and open context menu)

app/assets/javascripts/components/NotesListItem.tsx
```diff
@@ -0,0 +1,142 @@
...
+            <span>{note.createdAtString || 'Now'}</span>
+          ) : null}
+        </div>
+      ) : null}
+      {!hideTags ? (
+        <div className="tags-string">
+          <div className="faded">{tags}</div>
+        </div>
+      ) : null}
+    </div>
```
laurentsenta, line 140: I don't feel strongly about it, but you could use `&&` operator, we had a similar conversation on TagsList:

https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/components/TagsListItem.tsx#L103-L107

app/assets/javascripts/components/NotesView.tsx
```diff
@@ -0,0 +1,256 @@
...
+      application.io,
+      createNewNote,
+      searchBarElement,
+      selectNextNote,
+      selectPreviousNote,
+    ]);
+
+    const onNoteFilterTextChange = (e: Event) => {
+      setNoteFilterText((e.target as HTMLInputElement).value);
+    };
```
laurentsenta, line 118: For what it's worth,
there are types for these, they are tricky to find, but they make your code fully type checked,

https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/components/TagsListItem.tsx#L45

app/assets/stylesheets/_notes.scss
```diff
@@ -1,11 +1,16 @@
+notes-view {
+  width: 350px;
+}
+
```
laurentsenta, line 4: Should we use a class `.notes-view` here?

I really like how angular let you define new html element, but we'd lose the style without really knowing why when we migrate the parent component to react.

laurentsenta (commented):
That is some huge and tedious change, gg :)
A few questions regarding the code, but the migration looks neat

laurentsenta (approved):

PR #778, opened 2021-12-16
Beta/backups in prefs

vardan-arm:


app/assets/javascripts/preferences/panes/backups-segments/EmailBackups.tsx
```diff
@@ -0,0 +1,204 @@
...
+  };
+
+  const toggleMuteFailedCloudBackupEmails = async () => {
+    setIsFailedCloudBackupEmailMuted(!isFailedCloudBackupEmailMuted);
+
+    await updateUserSetting(
+      SettingName.MuteFailedCloudBackupsEmails,
+      `${!isFailedCloudBackupEmailMuted}`
+    );
+  };
```
laurentsenta, line 120: Would it be helpful to move the business logic to a MobX store?

It looks like the component deals with quite a lot of business-related logic (not just `useState` to hold the content of text input but also a lot of state related to async operation with the data, loading states, etc),

Usually that's the sign that a mobx store could help separate concerns (business vs interface) and make your react component simpler.

A few refs:

https://mobx.js.org/react-integration.html
https://developer.mozilla.org/en-US/docs/Glossary/MVC

laurentsenta (approved):

amanharwara (approved):

PR #951, opened 2022-03-24
fix: error on aborting file selection

amanharwara:
[Internal link](https://app.asana.com/0/1199942776723820/1201942428191885/f)


app/assets/javascripts/ui_models/app_state/files_state.ts
```diff
@@ -93,6 +93,10 @@ export class FilesState {
           ? await StreamingFileReader.getFilesFromHandles([fileOrHandle])
           : await picker.selectFiles();
 
+      if (!selectedFiles.length) {
```
moughxyz, line 96:
```suggestion
      if (selectedFiles.length === 0) {
```

Better to keep types consistent

PR #957, opened 2022-03-29
feat: new lock screen and challenge modal design

amanharwara:
[Internal link](https://app.asana.com/0/1201653402817568/1202006836122057/f)


app/assets/javascripts/components/ChallengeModal/ChallengePrompt.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    if (isInvalid) {
+      inputRef.current?.focus();
+    }
+  }, [isInvalid]);
+
+  return (
+    <>
+      {prompt.validation === ChallengeValidation.ProtectionSessionDuration ? (
+        <>
```
vardan-arm, line 44: I guess the top-level fragment is sufficient and we can skip fragments in this and below conditions.
```suggestion
```

app/assets/javascripts/components/ChallengeModal/ChallengeModal.tsx
```diff
@@ -0,0 +1,242 @@
...
+    }
+    const processingPrompts = valuesToProcess.map((v) => v.prompt);
+    setIsProcessing(processingPrompts.length > 0);
+    setProcessingPrompts(processingPrompts);
+    /**
+     * Unfortunately neccessary to wait 50ms so that the above setState call completely
+     * updates the UI to change processing state, before we enter into UI blocking operation
+     * (crypto key generation)
+     */
+    setTimeout(() => {
```
vardan-arm, line 95: Could we use promises here, which resolves after the above actions? That will let us not to depend from setTimeout and also will eliminate potential issues when 50 ms was not enough (because of heavy system load, for example).

app/assets/javascripts/components/Menu/Menu.tsx
```diff
@@ -117,7 +118,7 @@ export const Menu: FunctionComponent<MenuProps> = ({
       style={style}
       aria-label={a11yLabel}
     >
-      {Array.isArray(children) ? children.map(mapMenuItems) : null}
+      {toChildArray(children).map(mapMenuItems)}
```
vardan-arm, line 121: Will this work as prevously if there's only one child node? According to their (not very clear) [docs](https://preactjs.com/guide/v10/api-reference/#tochildarray) `toChildArray` will return an array with single element, but with our previous logic we return `null`.

PR #1047, opened 2022-05-24
fix: running tests and adding tests to CI & CD

karolsojko:


amanharwara (approved):

app/assets/javascripts/jest.config.js
```diff
@@ -25,4 +25,12 @@ module.exports = {
...
     '\\.svg$': 'svg-jest',
   },
+  coverageThreshold: {
+    global: {
+      branches: 3,
+      functions: 5,
+      lines: 21,
+      statements: 23,
+    },
+  },
```
vardan-arm, line 35: The positive numbers in "coverageThreshold" mean percentage, so maybe it makes more sense to increase these numbers (like to have branches coverage e.g. 60% instead of 3%)? Or the numbers are small in order to have the existing tests with low coverage passing right now?
https://jestjs.io/docs/configuration#coveragethreshold-object

vardan-arm (approved):

app/assets/javascripts/jest.config.js
karolsojko: Yes, the intention here is to define the threshold in a way that does not allow it to get any lower. So this is the "current state" of coverage and we don't want it to get worse :)

PR #1170, opened 2022-06-28
feat: move search bar to navigation panel

amanharwara:
[Internal link](https://app.asana.com/0/1201611182420182/1202480392174100/f)


moughxyz (approved):

jacobgkau:
What was the rationale for this change? I use search _instead of_ tags, and I keep the tags panel ("navigation panel") collapsed at all times to save space and remove clutter. Now I have to waste space with that panel or open it up every time I want to search.

PR #1392, opened 2022-08-10
fix: optimize toasts for mobile

amanharwara:


packages/toast/src/ToastTimer.tsx
```diff
@@ -55,14 +55,6 @@ export const ToastTimer: FunctionComponent<Props> = ({ toast, index }) => {
...
   }, [resumeTimer])
 
-  const handlePageVisibility = useCallback(() => {
-    if (document.visibilityState === 'hidden') {
-      pauseTimer()
-    } else {
-      resumeTimer()
-    }
-  }, [pauseTimer, resumeTimer])
-
```
vardan-arm, line 65: Why do we remove this function, is it no longer needed?

vardan-arm (approved):

PR #1907, opened 2022-10-31
fix(mobile): increase font sizes

moughxyz:


packages/web/src/javascripts/Components/ContentListView/Header/ContentListHeader.tsx
```diff
@@ -87,7 +87,7 @@ const ContentListHeader = ({
     return (
       <button
         className={classNames(
-          'absolute bottom-6 right-6 z-editor-title-bar ml-3 flex h-15 w-15 cursor-pointer items-center',
+          'fixed bottom-6 right-6 z-editor-title-bar ml-3 flex h-15 w-15 cursor-pointer items-center',
```
moughxyz, line 90: @amanharwara heads up I changed this to fixed because with the desktop app absolute would be shifted from the top on macOS 25px, so this button would be out of view.

PR #2085, opened 2022-12-03
refactor: separate selectAndUploadFile and uploadFile functions

amanharwara:


moughxyz:
@amanharwara as convention, I’ll treat all review requests as optional and non-blocking. To indicate you really want a review before merging, assign the PR to me on internal or otherwise specify that you’d like an explicit review. But otherwise you can treat no reply from me as abstaining due to lack of context/familiarity.

The reason I say this is I’m sometimes not sure if you’re waiting on me or not, so some kind of explicit flagging system might help clear that up.

PR #2125, opened 2022-12-29
sync desktop electron version

proletarius101:
It makes no sense to depend on electron on one version while build against another


moughxyz:
Hmm good catch, I think we could just remove this, and it should default to package version. https://www.electron.build/configuration/configuration.html

moughxyz:
Done via https://github.com/standardnotes/app/commit/a896b4d9cfcda1221aad1d847eec104efd8757ba

PR #2197, opened 2023-02-04
fix: gradle version inconsistency

proletarius101:
This leads to inconsistent gradle version detection, e.g. on F-Droid it is parsed as 7.2, while elsewhere it is probably 7.5.1.

It seems that `packages/mobile/android/app/gradle` is not used in the build process. We'd better remove it.


moughxyz:
Sorry, doesn't look like anyone was tagged or assigned here.

I'd be loath to merge this given that it's updates to a binary file we can't really see the contents of. Could you instead give us the location we can retrieve the same contents from, and we can make the change ourselves?

moughxyz:
Ah ok. Will merge this then once we take care of a few CI things related to mobile.

PR #2338, opened 2023-06-23
ContentListItem accessibility bug

cppshane:
When attempting to access Standard Notes via a screen reader, the list of notes for a given tag are not detected as being clickable elements because they are defined as `div` elements instead of `input` or `button` elements:

image

This PR just adds `role="button"` to them so that they are properly detected by screen readers as clickable elements.


amanharwara:
Hey @cppshane thanks for the PR! I reckon we should also make this change for [`FileListItem.tsx`](https://github.com/standardnotes/app/blob/main/packages/web/src/javascripts/Components/ContentListView/FileListItem.tsx)?

cppshane:
Good call, just added it to that as well 👍

amanharwara (approved):

PR #2371, opened 2023-07-26
fix: electron update

moughxyz:


moughxyz:
Postponing due to:

- https://bugs.chromium.org/p/chromium/issues/detail?id=1467110&q=css&can=2
- https://github.com/electron/electron/issues/37789

PR #2793, opened 2024-01-24
Bugfix: Simplenote converter handles empty notes correctly

mmorella-dev:
Fixes https://github.com/standardnotes/forum/issues/3462

The current implementation of SimplenoteConverter uses the following helper function to validate notes.json:
```ts
const isSimplenoteEntry = (entry: any): boolean => entry.id && entry.content && entry.creationDate && entry.lastModified
```

Because `&&` is used here, this function returns false when `content` is an empty string `""`. This can trivially occur in notes where the user clicked "New Note" but did not enter any data. For example:
```json
{
  "id": "e5ae4e85-a6a4-45d9-8994-6e693483f679",
  "content": "",
  "creationDate": "2024-01-02T01:37:22.615Z",
  "lastModified": "2024-01-02T01:38:26.662Z"
}
```

This PR rewrites the function to instead use `Object.prototype.hasOwnProperty` to for the existence of each property, which produces the expected output when `content` is an empty string.


amanharwara (changes requested):
Hi, thanks for the PR! 

Looks like lint is failing (https://github.com/standardnotes/app/actions/runs/7636015424/job/20817657669?pr=2793#step:5:57) Please fix those issues.

Also, please run `yarn test SimplenoteConverter` in `packages/ui-services` and make sure that passes.

mmorella-dev:
@amanharwara Thanks for the heads up! I was hasty to get this out last night so I only tested it in a REPL. The new version should avoid those Lint errors. Tests pass on my machine.

amanharwara (approved):
LGTM! Thanks again!

PR #2798, opened 2024-01-24
Bugfix: Fix Simplenote importer not titling multi-line documents

mmorella-dev:
SimplenoteConverter has a bug which causes it to only parse a title in notes which have *exactly* two lines. This is one of the issues addressed by [jamesgecko/simple-to-standard](https://github.com/jamesgecko/simple-to-standard).

Example data:
```json
{
  "activeNotes": [
    {
      "id": "bb33231fc77e49e9b2f66b5d07b1159d",
      "content": "Title One\r\nThis note has a title and one line of body!",
      "..."
    },
    {
      "id": "53c79c16-896f-43e5-80e5-a0f92c85330d",
      "content": "Title Two\r\nThis note has a title and...\r\nmultiple lines...\r\nof body!",
     "..."
    },
    {
      "id": "1b105f4d24534d329ca8d99439219685",
      "content": "\r\nThis note has only a body!",
      "..."
    },
  ],
  "..."
}
```

Current behavior: the second note is given a title.
Screenshot of Standard Notes running locally. The second note imported with the date as its title.

This PR fixes that issue by replacing an instance of String.prototype.split with a new helper function, splitOnce, which splits a string only on the first occurrence of the delimiter.

After fix: multi-line notes are titled correctly.
Screenshot of Standard Notes running locally. The second note imported with the correct title.

### DIFF
--- a/packages/ui-services/src/Import/SimplenoteConverter/SimplenoteConverter.ts
+++ b/packages/ui-services/src/Import/SimplenoteConverter/SimplenoteConverter.ts
@@ -15,6 +15,17 @@ type SimplenoteData = {
 const isSimplenoteEntry = (entry: any): boolean =>
   entry.id && entry.content != undefined && entry.creationDate && entry.lastModified
 
+const splitAtFirst = (str: string, delim: string): [string, string] | [] => {
+  const indexOfDelimiter = str.indexOf(delim)
+  const hasDelimiter = indexOfDelimiter > -1
+  if (!hasDelimiter) {
+    return []
+  }
+  const before = str.slice(0, indexOfDelimiter)
+  const after = str.slice(indexOfDelimiter + delim.length)
+  return [before, after]
+}
+
 export class SimplenoteConverter implements Converter {
   constructor() {}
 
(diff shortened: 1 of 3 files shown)


packages/ui-services/src/Import/SimplenoteConverter/SimplenoteConverter.ts
```diff
@@ -15,6 +15,15 @@ type SimplenoteData = {
 const isSimplenoteEntry = (entry: any): boolean =>
   entry.id && entry.content != undefined && entry.creationDate && entry.lastModified
 
+const splitOnce = (str: string, delim: string): string[] => {
+  const i = str.indexOf(delim)
+  if (i < 0) {
+    // no delimeter found, return whole string
```
amanharwara, line 21:
```suggestion
const splitOnce = (str: string, delim: string): string[] => {
  const indexOfDelimiter = str.indexOf(delim)
  const hasDelimiter = indexOfDelimiter > -1
  if (!hasDelimiter) {
```

Given that all of this code is going to be minified anyway, it doesn't make to use shortened variable names like `i` which don't give any context. It should probably be changed to something like `indexOfDelimiter`, along with `str`->`string` and `delim`->`delimiter`.

We always prefer using descriptive variable names instead of using comments. In this case, the conditional can be inverted and extracted into a const with a descriptive name like `hasDelimiter`.

amanharwara (changes requested):
Also, please add a test case to `SimplenoteConverter.spec.ts`. Hard to know the change of behavior without that.

mmorella-dev:
@amanharwara Thanks for the suggestions! I've made some stylistic changes in-line with your suggestions, and modified the test data to include a note with more than 2 lines. In my testing, this case fails on the old version (79951c388c8d81e0acb477b131cdfce66d040f34) and passes after my changes (a7f090f17a8bcca6912633f0728f4233e251bcbd).

(Sorry for the noise with all the force-pushes. I can get a bit carried away making history "perfect".)

mmorella-dev:
(okay, one more...)

mmorella-dev:
@amanharwara Build and tests passed!

amanharwara (approved):
LGTM, thanks!

PR #2871, opened 2024-04-20
feat: add fastlane app description

yurtpage:
The Fastlane allows to publish an app description for both PlayStore and F-Droid. The description can be internationalized.

Translation of the app description is very important because F-Droid doesn't even makes an auto translation.

Please integrate the app with the Weblate or Transifex and add the App Metadata to translation.
Here is a sample of Conversations app metadata translation via Weblate https://translate.codeberg.org/projects/conversations/app-store-metadata-conversations/

Meanwhile here I added one translation as a sample.


moughxyz:
Will these be used automatically when uploading to Google Play as well? I think for now I'd like to avoid that since we treat Google Play GUI as the source of truth for desc and screenshots.

yurtpage:
@moughxyz friendly remind on this, I signed the commit

PR #2882, opened 2024-08-30
Fix blurry right click menus

DarkArc:
This is ~~an untested~~ a change to resolve a problem I noticed in the desktop app and then corrected by adjusting the CSS in the inspector.

The font blurriness seems to only occur when the pixels are odd numbers, thus by using even numbers we can avoid the blurry right click menu that sometimes appears.

### DIFF
--- a/packages/web/src/javascripts/Components/Popover/GetPositionedPopoverStyles.ts
+++ b/packages/web/src/javascripts/Components/Popover/GetPositionedPopoverStyles.ts
@@ -78,12 +78,24 @@ const getStylesFromRect = (options: {
   const shouldApplyMobileWidth = isMobileScreen() && disableMobileFullscreenTakeover && !disableApplyingMobileWidth
   const marginForMobile = percentOf(10, window.innerWidth)
 
+  let xTranslation = shouldApplyMobileWidth ? marginForMobile / 2 : Math.floor(rect.x)
+  let yTranslation = Math.floor(rect.y)
+
+  // There is a bug in Chrome that results in blurry results from translate calls:
+  //
+  //   https://stackoverflow.com/questions/32034574/font-looks-blurry-after-translate-in-chrome
+  //
+  // To work around this issue and ensure the right click menu is always pixel perfect,
+  // ensure even numbers are used.
+  xTranslation = Math.floor(xTranslation / 2) * 2
+  yTranslation = Math.floor(yTranslation / 2) * 2
+
   return {
     willChange: 'transform',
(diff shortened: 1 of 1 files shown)


packages/web/src/javascripts/Components/Popover/GetPositionedPopoverStyles.ts
```diff
@@ -78,12 +78,24 @@ const getStylesFromRect = (options: {
...
+
   return {
     willChange: 'transform',
-    '--translate-x': `${shouldApplyMobileWidth ? marginForMobile / 2 : Math.floor(rect.x)}px`,
-    '--translate-y': `${Math.floor(rect.y)}px`,
+    '--translate-x': `${xTranslation}px`,
+    '--translate-y': `${yTranslation}px`,
     '--offset': `${options.offset}px`,
-    transform: 'translate3d(var(--translate-x), var(--translate-y), 0)',
+    transform: 'translate(var(--translate-x), var(--translate-y))',
```
DarkArc, line 98: This might be wrong, but `translate(x, y)` seemed to be just as effective as `translate3d(x, y, 0)`

DarkArc:
Before:
image
After:
image

DarkArc:
This is now tested.

moughxyz:
How can we replicate this? Changing the translate-x/y value in inspector to an odd number doesn't yield a blurry menu.

DarkArc:
That's interesting; I see this on two devices when using the desktop app. Both devices are running Linux and using the standard notes desktop app flatpak.

The device with the 4k monitor the effect is more subtle. The other device is a framework laptop. 

One thought is it might have something to do with display scaling factors in chromium under Linux.

moughxyz:
Can you share a screen recording?

DarkArc:
I'm not really sure what that would do... It's exactly as I described and the screenshots show above.

DarkArc:
I really think this was a waste of (my) time but here's the requested video.
https://github.com/user-attachments/assets/6d5b8bc0-08b0-4053-b3c8-88db279415b7

DarkArc:
Here's a better one (higher resolution) 
https://github.com/user-attachments/assets/55272db0-0c4f-4101-b61b-9043abaf0192

Entrpi:
Eager to see this merged, very distracting issue.

antsgar:
@DarkArc I know this has been open for a really long time, just wanted to let you know I'm including it in the next release. Apologies for the long wait, and thank you for contributing!

PR #2887, opened 2024-10-25
fix: toggle notes panel

luis-411:
This fixes the issue mentioned here: https://github.com/standardnotes/forum/issues/3754

Somehow `itemsPanelResizeWidthChangeCallback` returns 0 and so `itemsPanelWidth` is also 0 and therefore the panel is not shown.
This could be due to an error in the `ContentListView.tsx` file, which I was not able to find.

However, I added this kind of workaround so now it works flawlessly.
If somebody knows a better solution, please let me know and I will be happy to change it accordingly.

### DIFF
--- a/packages/web/src/javascripts/Components/Panes/PanesSystemComponent.tsx
+++ b/packages/web/src/javascripts/Components/Panes/PanesSystemComponent.tsx
@@ -207,7 +207,7 @@ const PanesSystemComponent = () => {
             }
           } else {
             return {
-              gridTemplateColumns: `${itemsPanelWidth}px auto`,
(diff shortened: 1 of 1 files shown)


luis-411:
Dear @karolsojko , @moughxyz , @amanharwara , @effieeee,

I think this pull request can fix a relevant issue. Could you please help me review it?

Thanks,
Luis

amanharwara:
This doesn't fix the actual underlying issue, but rather just covers it.

Looked at it and it was an issue with refs, fixed by https://github.com/standardnotes/app/commit/7516aae885f3ba7512c1da44212ea9afb8036550

Thanks for the PR regardless!

PR #2891, opened 2025-03-05
First commit to add a Modal about Linked notes

diegorv:
Hey team!

This is just a suggestion. Since it's my first time interacting with the app's source code, please feel free to conduct the code review as needed.

Current Behavior: When viewing a note, in the note header, we currently display only 5 linked tags/notes. After that, it shows "and more X...".
CleanShot 2025-03-05 at 13 20 23

New Behavior: When viewing a note, in the note header, we display only the first 5 linked tags/notes as before. After that, we show a button labeled "(total count) more". When clicking the button, a modal opens displaying all the notes.
CleanShot 2025-03-05 at 13 20 43

CleanShot 2025-03-05 at 13 24 17

Questions:
- Should I create a Component.test.tsx file to test each of these changes? I'm asking because I couldn't find specific tests for LinkedItemBubblesContainer, and I'm a bit confused about what I should do.
- If you could clarify how you would like me to write the tests, along with any references or guidelines, that would be great!

I'm loving your application! Congratulations on the great work! I even made sure to subscribe to the annual professional plan :)


amanharwara:
Hey, thanks for the PR.

The code itself looks good to me, but what would be the benefit of adding this when there is already the existing Linked Items button? The existing one seems better, it groups the linked items according to type and also allows to search items that are already linked and also ones that you can link.

image

PR #2900, opened 2025-05-16
opt-in to MTE async mode

cameronaaron:
currently only benefits GrapheneOS users on Pixel 8 and 9 who don't turn on MTE for all apps or Molly manually. hopefully will be extended to all Pixel users and then more devices


amanharwara:
could you add more context on what this does?

cameronaaron:
This change improves the app’s ability to catch memory safety bugs on devices that support ARM’s Memory Tagging Extension, with minimal performance impact. It’s a proactive step toward more robust and secure code, especially important for a widely-used app it will also allow the app to take advantage of the mem tagging in Android 16

cameronaaron:
I added a similar patch to mullvad that was merged in

TheAutisticTechie:
This would only catch bugs in production. Sync mode needs to be set up for the test environment first as per recommended usage: https://source.android.com/docs/security/test/memory-safety/arm-mte#recommended-use

PR #2906, opened 2025-06-22
feat: Tag search will now only show the direct tag results instead of also showing the parent and siblings

amanharwara:
- Refactors tag search to only show the direct relevant tags and not any parent, siblings or children
- Double-clicking a tag when searching will clear search and scroll the tag into view.

https://github.com/user-attachments/assets/1ca68d95-141e-4815-9f8d-c26ab98a6f6b

### DIFF
--- a/packages/web/src/javascripts/Components/Tags/TagsList.tsx
+++ b/packages/web/src/javascripts/Components/Tags/TagsList.tsx
@@ -6,16 +6,26 @@ import { TagListSectionType } from './TagListSection'
 import { TagsListItem } from './TagsListItem'
 import { useApplication } from '../ApplicationProvider'
 import { useListKeyboardNavigation } from '@/Hooks/useListKeyboardNavigation'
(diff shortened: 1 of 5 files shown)


packages/web/src/javascripts/Components/Tags/TagsList.tsx
```diff
@@ -6,16 +6,26 @@ import { TagListSectionType } from './TagListSection'
 import { TagsListItem } from './TagsListItem'
 import { useApplication } from '../ApplicationProvider'
 import { useListKeyboardNavigation } from '@/Hooks/useListKeyboardNavigation'
+import { NavigationController } from '@/Controllers/Navigation/NavigationController'
 
 type Props = {
   type: TagListSectionType
 }
 
+function getAllTagsForType(controller: NavigationController, type: TagListSectionType) {
```
antsgar, line 15: I'd suggest either just passing `isSearching` here as a parameter instead of the entire controller object, or just have this function read from `application`, that way it'd be clearer to see what input the function needs

PR #2907, opened 2025-06-23
fix(SNOTES-468): use injectJavascript to set safe area inset vars

antsgar:
I've only been able to repro this on my production app by leaving the app open (minimized) overnight and then focusing on it again the next day. I'm not entirely sure what's going on, but I suspect that in those conditions either the effect that triggers the message to update the insets isn't fired, or the condition that checks for `didAppLaunch` before sending the message returns false.

This PR sets the CSS safe area inset vars using `injectJavascript` instead of `postMessage`. Injecting JS is actually the recommended way to pass information from RN to Web per the react-native-webview docs, and there's an example in the docs where this is used to update the background color of the webview https://github.com/react-native-webview/react-native-webview/blob/master/docs/Guide.md#the-injectjavascript-method. This will run on every render, and that way we ensure that the correct inset paddings are always being set.

### DIFF
--- a/packages/mobile/src/MobileWebAppContainer.tsx
+++ b/packages/mobile/src/MobileWebAppContainer.tsx
@@ -458,7 +458,7 @@ const MobileWebAppContents = ({ destroyAndReload }: { destroyAndReload: () => vo
             component: CustomAndroidWebView,
           } as WebViewNativeConfig,
         })}
-        webviewDebuggingEnabled={true}
+        webviewDebuggingEnabled
       />
     </View>
   )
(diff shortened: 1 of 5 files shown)


packages/mobile/src/MobileWebAppContainer.tsx
```diff
@@ -136,7 +137,7 @@ const MobileWebAppContents = ({ destroyAndReload }: { destroyAndReload: () => vo
       keyboardDidHideListener.remove()
       keyboardWillChangeFrame.remove()
     }
-  }, [webViewRef, stateService, device, androidBackHandlerService, colorSchemeService])
+  }, [webViewRef, stateService, device, androidBackHandlerService, colorSchemeService, insets.bottom])
```
antsgar, line 140: Unrelated change, this was a missing dependency for this useEffect

PR #2915, opened 2025-07-17
fix: Remove extra bottom padding on Android

antsgar:


packages/web/src/javascripts/Hooks/useSafeAreaPadding.ts
```diff
@@ -14,9 +14,9 @@ export const useAvailableSafeAreaPadding = () => {
   const [isKeyboardVisible, setIsKeyboardVisible] = useState(false)
   useEffect(() => {
     return application.addNativeMobileEventListener((event) => {
-      if (event === ReactNativeToWebEvent.KeyboardWillShow) {
+      if (event === ReactNativeToWebEvent.KeyboardWillShow || event === ReactNativeToWebEvent.KeyboardDidShow) {
         setIsKeyboardVisible(true)
-      } else if (event === ReactNativeToWebEvent.KeyboardWillHide) {
+      } else if (event === ReactNativeToWebEvent.KeyboardWillHide || event === ReactNativeToWebEvent.KeyboardDidHide) {
```
antsgar, line 19: The change to which events determine the keyboard visibility from the last PR seem to have affected the keyboard visibility detection for Android, so adding back the previous events so it works for both platforms

PR #2916, opened 2025-07-18
chore: Improve release workflows

antsgar:


.github/workflows/desktop.release.reuse.yml
```diff
@@ -78,21 +78,25 @@ jobs:
...
-
-      - run: yarn build:desktop
-
       - run: echo APP_VERSION=$(node -p "require('./../web/package.json').version") >> $GITHUB_ENV
-
-      - name: Compile for AppImage
-        run: yarn run webpack --config desktop.webpack.prod.js
       - name: AppImage
-        run: yarn run electron-builder --linux --x64 -c.linux.target=AppImage --publish=never --c.extraMetadata.version=${{ env.APP_VERSION }}
+        run: |
```
antsgar, line 83: The native modules are currently being rebuilt once before the packaging for both architectures happens. I suspect that might be the reason the native modules are always x64 and the arm64 Linux packages aren't running properly (see https://github.com/standardnotes/forum/issues/3457)

.github/workflows/desktop.release.reuse.yml
```diff
@@ -78,21 +78,25 @@ jobs:
 
       - run: yarn install --immutable
 
-      - name: Rebuild Electron Native Modules
-        run: yarn workspace @standardnotes/desktop rebuild:home-server
-
-      - run: yarn build:desktop
```
antsgar, line 84: This is the same as `yarn run webpack --config desktop.webpack.prod.js` which runs below I think?

.github/workflows/desktop.release.reuse.yml
```diff
@@ -78,21 +78,25 @@ jobs:
...
-
       - run: echo APP_VERSION=$(node -p "require('./../web/package.json').version") >> $GITHUB_ENV
-
-      - name: Compile for AppImage
```
antsgar, line 88: Now done in each step after rebuilding

packages/desktop/package.json
```diff
@@ -55,7 +55,7 @@
     "@babel/core": "*",
     "@babel/preset-env": "*",
     "@electron/notarize": "^3.0.0",
-    "@electron/rebuild": "3.6.0",
+    "@electron/rebuild": "3.6.1",
```
antsgar, line 58: There's a fix on 3.6.1 which might affect rebuilding for different platforms https://github.com/electron/rebuild/pull/1153

amanharwara (approved):
LGTM at a glance. Only way to test this afaik is to merge and trigger a new release

PR #2941, opened 2025-09-26
fix: Fixes pasting from Google Docs always uses bold font

antsgar:
When copying text from Google Docs currently it pastes on SN using bold font. This is due to Google Docs misusing a `<b>` tag for all text and then using inline styling to set the actual font weight. Since the editor is rendered within an iframe, inline styling is disallowed and not parsed correctly, which results in the `<b>` tag rendering bold text. This PR adds a custom plugin to detect when the pasted text comes from Google Docs and removes the `<b>` tag. Note that this will only get rid of the bold font, but won't solve the inline styling not being applied (eventually we could evolve this plugin to also handle inline styling and translate it to semantic tags).

Context on Google Docs misuse of HTML tags: https://github.com/froala/wysiwyg-editor/issues/1918


amanharwara:
Can you check if the issue also happens on https://playground.lexical.dev? If it doesn't then that means this has already been fixed upstream and we just need to bump versions. If not, we can merge this for now but also try to open a PR with a fix for upstream. We already have a [forked repo](https://github.com/standardnotes/lexical) which you can use to create a PR if needed.

PR #2999, opened 2026-03-28
Fix the default language style for newly created code blocks

micl2e2:
It seems that most code in `CodeOptions.tsx` that tries to set default language as 'empty' provides little help, and somehow always fallback to JS (https://github.com/facebook/lexical/blob/5dd93a7c8b4b23d2b00a547c2d860cb17aed1e52/packages/lexical-code-core/src/CodeNode.ts#L52).

I marked this as a "fix" is because line 18 suggests that the plain text type was needed at the very beginning.


micl2e2:
May I ask, as paying customers, specifically "Professional" ones, how long could we expect our contributions to get responded from the team?

antsgar:
Hi @micl2e2! Thank you for your contribution, changes look good. Apologies for the delay in reviewing your PR 🙏 we try to review all contributor PRs as soon as possible, but response times can vary depending on the team's current priorities and workload. We appreciate your patience and your help.

PR #3014, opened 2026-06-02
POC - multi tab notes

TheMany172:
image


TheMany172:
It is a bit messy of course, and I did it in windows, which appears to not be the dev preference (nor mine...)

but it is just there to show that it is completely possible to have multi tabs

and split screen editing

PR #3034, opened 2026-07-31
fix: preferences sidebar unclickable when opened in focus mode

### LINKED ISSUE #4140: Preferences sidebar cannot be clicked when opening settings while in focus mode
The preferences sidebar/navbar rejects all clicks when the preferences page is opened while in a note with focus mode enabled and the application window width is greater than ~1,200 px. In this case, the preferences pages **cannot be navigated**.

**To Reproduce**
1. Make sure application window is at least 1,200 px wide
1. Open a note
1. Activate focus mode
1. Open the preferences page via hotkey (e.g. `⌘ + ,` on macOS)
1. Attempt to navigate between preferences pages using the sidebar

**Expected behavior**
The preferences page navbar items should activate when clicked on, allowing the user to navigate between preferences pages.

**Desktop:**
 - OS: macOS 26.5
 - Desktop app
 - Version 3.201.21

**Additional context**
Observed on a 4K external monitor.

jayeesha:
In focus mode at viewport widths ≥ 1200px, #editor-column padding grows to 20% and is marked -webkit-app-region: drag. macOS hit-tests drag regions above the web layer, so this padding band swallows mouse events for the preferences sidebar rendered beneath it — the menu items are plain divs, not button/input elements exempted globally in renderer.css.

Fixed by marking #editor-column as no-drag in focus mode. Window dragging is unaffected since the #editor-column:before pseudo-element already provides a dedicated drag area.

Tested with focus mode on/off across viewport widths (~900, ~1100, ~1300, ~1600px), confirming preferences sidebar works in all cases and window dragging still works via the top strip.

Fixes standardnotes/forum#4140

### DIFF
--- a/packages/desktop/app/stylesheets/renderer.css
+++ b/packages/desktop/app/stylesheets/renderer.css
@@ -87,6 +87,10 @@
   -webkit-app-region: no-drag;
 }
 
+.mac-desktop.focus-mode #editor-column {
+  -webkit-app-region: no-drag;
+}
+
 .mac-desktop [data-dialog-portal]:has([data-dialog])::after {
   content: '';
   position: absolute;


antsgar:
@jayeesha thank you for your contribution! Merging this one in for the next release.

